PR #6, opened 2019-05-09
Manage status

enxebre:
- Add status API types
- Create proper struct type <-> rawExtension methods and tests for providerSpec and providerStatus
- Add providerID and providerStatus to machine scope and use new struct type <-> rawExtension methods
- Add reconcileMachineWithCloudState() to be called on Create for now


pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -35,6 +36,7 @@ func newReconciler(scope *machineScope) *Reconciler {
 
 // Create creates machine if and only if machine exists, handled by cluster-api
 func (r *Reconciler) create() error {
+	defer r.reconcileMachineWithCloudState()
```
frobware: Can create be called multiple times for `r`? If so, what happens with the multiple `defer` calls?

PR #8, opened 2019-05-09
Validate earlier

michaelgugino:


enxebre:
let's hold on this one until we have some use cases
/hold

PR #9, opened 2019-05-10
Implement Delete instance

michaelgugino:


pkg/cloud/gcp/actuators/machine/actuator.go
```diff
@@ -11,6 +11,10 @@ import (
 	controllerclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const (
+	scopeFailMsg = "failed to create scope for machine %q: %v"
```
frobware: I think this would be better as `scopeFailFmt` - then calls to `fmt.Errorf(scopeFailFmt, ...)` would read a lot better.

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -195,3 +198,63 @@ func validateMachine(machine machinev1.Machine, providerSpec v1beta1.GCPMachineP
...
+	_, err := r.computeService.InstancesGet(r.projectID, zone, r.machine.Name)
+	if err == nil {
+		return true, nil
+	}
+	if isNotFoundError(err) {
+		return false, nil
+	}
+
+	return false, fmt.Errorf("Error getting running instances: %v", err)
+
```
frobware: nit; drop blank line.

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -195,3 +198,63 @@ func validateMachine(machine machinev1.Machine, providerSpec v1beta1.GCPMachineP
...
+	}
+	zone := r.providerSpec.Zone
+	operation, err := r.computeService.InstancesDelete(r.projectID, zone, r.machine.Name)
+	if err != nil {
+		return fmt.Errorf("failed to delete instance via compute service: %v", err)
+	}
+	if op, err := r.waitUntilOperationCompleted(zone, operation.Name); err != nil {
+		return fmt.Errorf("failed to wait for delete operation via compute service. Operation status: %v. Error: %v", op, err)
+	}
+	if err != nil {
```
frobware: Why do we need this line? If calls to `InstanceDelete()` fail we return (line 238). If calls to `waitUntilOperationCompleted()` fail we return (line 241). At this point delete has succeeded.

pkg/cloud/gcp/actuators/machine/reconciler_test.go
```diff
@@ -89,3 +87,44 @@ func TestReconcileMachineWithCloudState(t *testing.T) {
...
+				Namespace: "",
+			},
+		},
+		coreClient:     controllerfake.NewFakeClient(),
+		providerSpec:   &gcpv1beta1.GCPMachineProviderSpec{},
+		providerStatus: &gcpv1beta1.GCPMachineProviderStatus{},
+		computeService: mockComputeService,
+	}
+	reconciler := newReconciler(&machineScope)
+	if err := reconciler.delete(); err != nil {
```
frobware: Can we add a test case where delete is expected to fail?

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -21,6 +21,8 @@ const (
 	userDataSecretKey  = "userData"
 	operationTimeOut   = 180 * time.Second
 	operationRetryWait = 5 * time.Second
+	gceTimeout         = time.Minute * 10
```
enxebre: where is this used?

pkg/cloud/gcp/actuators/machine/reconciler.go
michaelgugino: @frobware this is based on upstream implementation and also create.  While the API request my have been successful, it looks like there's some kind of polling mechanism in the background to ensure that the request actually takes place.  Delete and create are async.

michaelgugino: nvm, I see I left an extra if statement in there.

enxebre:
can we drop `Implement Exists` commit?

pkg/cloud/gcp/actuators/machine/actuator.go
```diff
@@ -82,6 +86,15 @@ func (a *Actuator) Update(ctx context.Context, cluster *clusterv1.Cluster, machi
...
+	klog.Infof("Deleting machine %v", machine.Name)
+	scope, err := newMachineScope(machineScopeParams{
+		machineClient: a.machineClient,
+		coreClient:    a.coreClient,
+		machine:       machine,
+	})
+	if err != nil {
+		return fmt.Errorf(scopeFailFmt, machine.Name, err)
+	}
+	defer scope.Close()
```
enxebre: we don't need to close and store here

PR #14, opened 2019-05-13
Status conditions

enxebre:
Creation error:
```
status:
  providerStatus:
    conditions:
    - lastProbeTime: "2019-05-13T17:58:12Z"
      lastTransitionTime: "2019-05-13T17:58:12Z"
      message: 'googleapi: Error 400: Invalid value for field ''resource.disks[0].initializeParams.diskType'':
        ''zones/us-east1-b/diskTypes/pd-standardasdasd''. The referenced diskType
        resource cannot be found., invalid'
      reason: MachineCreationFailed
      status: "False"
      type: MachineCreated
    metadata:
      creationTimestamp: null
```

Creation success:
```
status:
  addresses:
  - address: 10.0.0.40
    type: InternalIP
  - address: 35.237.91.4
    type: ExternalIP
  providerStatus:
    conditions:
    - lastProbeTime: "2019-05-13T18:05:22Z"
      lastTransitionTime: "2019-05-13T18:05:22Z"
      message: machine successfully created
      reason: MachineCreationSucceeded
      status: "True"
      type: MachineCreated
    instanceId: agl060519-rnvnt-test-gcp-34
    instanceState: RUNNING
    metadata:
      creationTimestamp: null
```


pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -119,39 +119,70 @@ func (r *Reconciler) create() error {
...
+			Reason:  machineCreationFailedReason,
+			Message: err.Error(),
+			Status:  corev1.ConditionFalse,
+		}); reconcileWithCloudError != nil {
+			klog.Errorf("Failed to reconcile machine with cloud state: %v", reconcileWithCloudError)
+		}
 		return fmt.Errorf("failed to wait for create operation via compute service. Operation status: %v. Error: %v", op, err)
 	}
-	return nil
+	return r.reconcileMachineWithCloudState(nil)
```
michaelgugino: This has an api call that could potentially fail, even though the instance was created (same with a timeout of waitUntilOperationCompleted above).  We need to add logic to check that a machine was already created (instance.insert), so we don't fail on trying to create it again, rinse repeat.

enxebre: If any of those fail then the machine controller triggers a new reconciling iteration, then:
- The machine controller calls Exists() the actuator returns true, then:
- The machine controller calls update() the actuator reconciles state with cloud
This is by design.

PR #17, opened 2019-05-15
Implement events

michaelgugino:


pkg/cloud/gcp/actuators/machine/actuator.go
```diff
@@ -47,10 +68,17 @@ func (a *Actuator) Create(ctx context.Context, cluster *clusterv1.Cluster, machi
 		machine:       machine,
 	})
 	if err != nil {
-		return fmt.Errorf(scopeFailFmt, machine.Name, err)
+		fmtErr := fmt.Errorf(scopeFailFmt, machine.Name, err)
+		return a.handleMachineError(machine, apierrors.InvalidMachineConfiguration("error decoding MachineProviderConfig: %v", fmtErr), createEventAction)
```
enxebre: why is this passing using `InvalidMachineConfiguration`
and `error decoding MachineProviderConfig`?? 
Is there any value on handleMachineError? wouldn't it be simpler just `eventf` here - Same question for all the functions

pkg/cloud/gcp/actuators/machine/actuator.go
```diff
@@ -47,10 +68,17 @@ func (a *Actuator) Create(ctx context.Context, cluster *clusterv1.Cluster, machi
...
 	}
 	defer scope.Close()
-	return newReconciler(scope).create()
+	if err := newReconciler(scope).create(); err != nil {
+		return a.handleMachineError(machine, apierrors.InvalidMachineConfiguration("error creating Machine: %v", err), createEventAction)
+	}
+	// This event is best-effort and might get missed in case of timeout
+	// on waitUntilOperationCompleted
+	a.eventRecorder.Eventf(machine, corev1.EventTypeNormal, "Created", "Created Machine %v", machine.Name)
```
enxebre: if we create constants why don't use them https://github.com/openshift/cluster-api-provider-gcp/pull/17/files#diff-b314f951e6142d0ce7df84e1af378d85R22

pkg/cloud/gcp/actuators/machine/actuator_test.go
```diff
@@ -0,0 +1,33 @@
...
+		Events: eventsChannel,
+	}
+	cs := capifake.NewSimpleClientset()
+	// Initialize machine actuator.
+	machineActuator := NewActuator(ActuatorParams{
+		MachineClient: cs.MachineV1beta1(),
+		CoreClient:    fake.NewFakeClient(),
+		EventRecorder: recorder,
+	})
+	if machineActuator == nil {
```
enxebre: the way it's right now In a follow up we can may be allow to create a secret with credentials, which will be used to create a computeService which won't work and so we validate failure event.
Then if we want to validate success we'd need to allow `computeservice` injection at this level

frobware:
I also think the location where events are emitted should be at source. So in the implementation calls to (lowercase) create, delete, et al and not in the wrappers calls. That way you could [probably] dispense with "noEventAction" too.

pkg/cloud/gcp/actuators/machine/actuator.go
ingvagabund: Since `Actuator.handleMachineError` logs error and records the same error as event at the same time.

ingvagabund: `s/error decoding MachineProviderConfig/error creating machine scope` would be more appropriate

pkg/cloud/gcp/actuators/machine/actuator.go
```diff
@@ -79,10 +107,17 @@ func (a *Actuator) Update(ctx context.Context, cluster *clusterv1.Cluster, machi
 		machine:       machine,
 	})
 	if err != nil {
-		return fmt.Errorf("failed to create scope for machine %q: %v", machine.Name, err)
+		fmtErr := fmt.Errorf(scopeFailFmt, machine.Name, err)
+		return a.handleMachineError(machine, apierrors.InvalidMachineConfiguration("error decoding MachineProviderConfig: %v", fmtErr), updateEventAction)
```
ingvagabund: `s/error decoding MachineProviderConfig/error creating machine scope` would be more appropriate

pkg/cloud/gcp/actuators/machine/actuator.go
```diff
@@ -93,7 +128,12 @@ func (a *Actuator) Delete(ctx context.Context, cluster *clusterv1.Cluster, machi
 		machine:       machine,
 	})
 	if err != nil {
-		return fmt.Errorf(scopeFailFmt, machine.Name, err)
+		fmtErr := fmt.Errorf(scopeFailFmt, machine.Name, err)
+		return a.handleMachineError(machine, apierrors.InvalidMachineConfiguration("error decoding MachineProviderConfig: %v", fmtErr), deleteEventAction)
```
ingvagabund: `s/error decoding MachineProviderConfig/error creating machine scope` would be more appropriate

pkg/cloud/gcp/actuators/machine/actuator.go
enxebre: `InvalidMachineConfiguration` and `error decoding MachineProviderConfig` are not appropriate here

pkg/cloud/gcp/actuators/machine/actuator.go
```diff
@@ -47,10 +68,17 @@ func (a *Actuator) Create(ctx context.Context, cluster *clusterv1.Cluster, machi
...
+		fmtErr := fmt.Errorf(scopeFailFmt, machine.Name, err)
+		return a.handleMachineError(machine, apierrors.InvalidMachineConfiguration("error decoding MachineProviderConfig: %v", fmtErr), createEventAction)
 	}
 	defer scope.Close()
-	return newReconciler(scope).create()
+	if err := newReconciler(scope).create(); err != nil {
+		return a.handleMachineError(machine, apierrors.InvalidMachineConfiguration("error creating Machine: %v", err), createEventAction)
+	}
+	// This event is best-effort and might get missed in case of timeout
+	// on waitUntilOperationCompleted
```
enxebre: do we still use `waitUntilOperationCompleted`?

pkg/cloud/gcp/actuators/machine/actuator.go
enxebre: `InvalidMachineConfiguration` does not seem appropriate here

pkg/cloud/gcp/actuators/machine/actuator.go
enxebre: InvalidMachineConfiguration does not seem appropriate here

pkg/cloud/gcp/actuators/machine/actuator.go
```diff
@@ -79,10 +107,17 @@ func (a *Actuator) Update(ctx context.Context, cluster *clusterv1.Cluster, machi
...
+		fmtErr := fmt.Errorf(scopeFailFmt, machine.Name, err)
+		return a.handleMachineError(machine, apierrors.InvalidMachineConfiguration("error decoding MachineProviderConfig: %v", fmtErr), updateEventAction)
 	}
 	defer scope.Close()
-	return newReconciler(scope).update()
+	if err := newReconciler(scope).update(); err != nil {
+		return a.handleMachineError(machine, apierrors.InvalidMachineConfiguration("error decoding MachineProviderConfig: %v", err), updateEventAction)
+	}
+	// This event is best-effort and might get missed in case of timeout
+	// on waitUntilOperationCompleted
```
enxebre: is `waitUntilOperationCompleted` called here?

pkg/cloud/gcp/actuators/machine/actuator.go
enxebre: It's `waitUntilOperationCompleted` still used?

cmd/manager/main.go
```diff
@@ -36,6 +36,7 @@ func main() {
 	machineActuator := machine.NewActuator(machine.ActuatorParams{
 		MachineClient: cs.MachineV1beta1(),
 		CoreClient:    mgr.GetClient(),
+		EventRecorder: mgr.GetRecorder("gcp-controller"),
```
ingvagabund: After merging https://github.com/openshift/cluster-api-provider-gcp/pull/26 the code needs to be changed:

`s/mgr.GetRecorder("gcp-controller")/mgr.GetEventRecorderFor("gcpcontroller")`

pkg/cloud/gcp/actuators/machine/actuator.go
```diff
@@ -47,10 +68,15 @@ func (a *Actuator) Create(ctx context.Context, cluster *clusterv1.Cluster, machi
 		machine:       machine,
 	})
 	if err != nil {
-		return fmt.Errorf(scopeFailFmt, machine.Name, err)
+		fmtErr := fmt.Sprintf(scopeFailFmt, machine.Name, err)
+		return a.handleMachineError(machine, apierrors.InvalidMachineConfiguration(fmtErr), createEventAction)
```
enxebre: `apierrors.CreateMachine`?

ingvagabund: `InvalidMachineConfiguration` is generic error type (see https://github.com/kubernetes-sigs/cluster-api/blob/master/pkg/errors/machines.go#L42). Also see https://github.com/kubernetes-sigs/cluster-api/blob/master/pkg/apis/cluster/common/consts.go#L24-L29:
```
// Represents that the combination of configuration in the MachineSpec
// is not supported by this cluster. This is not a transient error, but
// indicates a state that must be fixed before progress can be made.
//
// Example: the ProviderSpec specifies an instance type that doesn't exist,
InvalidConfigurationMachineError MachineStatusError = "InvalidConfiguration"
```

`newMachineScope` might fail due to internal errors. Though, configuration is quite generic and not just about machine configuration. It's fine to consider even internal failures caused in invalid configuration.

enxebre: regardless what the comment says, that would print "InvalidConfiguration" which seems misleading to me. I would expect "InvalidConfiguration" to be used by a validate() func that as the comment says check that e.g instance type that doesn't exist. We are not forced to be using apierrors.* if they don't fit. IMHO All the abstractions here obfuscate what's actually being printed to the user for the sake of not duplicating a few lines.
This is not a blocker to me though.

michaelgugino: Personally, I don't think there's much utility in capture negative events.  99/100 they will be transient.  We should probably only record successful events as they're the only things that are useful with our current setup.  InvalidConfiguration I think is pretty useful here though; we're unlikely to recover if we fail to get a valid scope except for the infrequent case of GCP API outage.

ingvagabund: It's also fine to just use `return a.handleMachineError(machine, apierrors.CreateMachine("failed to create scope: %v", err), createEventAction)`

PR #18, opened 2019-05-15
[WIP] Simplify crud operations

frobware:
This removes a lot of the indirection through machineScope and helps to clean up the `defer` blocks that indirectly need to call update (on etcd).

TODO: Redo unit tests.


pkg/cloud/gcp/actuators/machine/actuator.go
```diff
@@ -41,59 +57,154 @@ func NewActuator(params ActuatorParams) *Actuator {
...
-		machine:       machine,
-	})
+
+	machineCtx, err := newMachineContext(a.coreClient, machine)
 	if err != nil {
-		return fmt.Errorf(scopeFailFmt, machine.Name, err)
+		return err
+	}
+
+	instance, err := instanceGet(machineCtx, machine.Name)
```
michaelgugino: This might be redundant as we call Exists() in the controller before we call Create().

pkg/cloud/gcp/actuators/machine/actuator.go
```diff
@@ -41,59 +57,154 @@ func NewActuator(params ActuatorParams) *Actuator {
...
-		machine:       machine,
-	})
+	klog.Infof("Checking if machine %q exists and is in RUNNING state", machine.Name)
+
+	machineCtx, err := newMachineContext(a.coreClient, machine)
+	if err != nil {
+		return false, err
+	}
+
+	if err := a.validateProviderSpec(machineCtx.providerSpec); err != nil {
```
michaelgugino: Maybe this should be part of newMachineContext ?

enxebre:
can we please put a separate scoped PR to make create/deletion async?

frobware:
> can we please put a separate scoped PR to make create/deletion async?

Why can't we finish this one up? We need to add back the unit tests.

pkg/cloud/gcp/actuators/machine/actuator.go
```diff
@@ -41,59 +57,154 @@ func NewActuator(params ActuatorParams) *Actuator {
...
+		instance = instanceFromMachineContext(machineCtx, machine.Name)
+		operation, err := machineCtx.computeService.InstancesInsert(machineCtx.projectID, machineCtx.providerSpec.Zone, instance)
+		if err != nil {
+			return err
+		}
+		klog.Infof("Creating instance %s/%s/%s using operation #%v", machineCtx.projectID, machineCtx.providerSpec.Zone, machine.Name, operation.Id)
+		// this forces further controller Reconciliation
+		machine.Spec.ProviderID = &machineCtx.providerID
+		_, err = a.machineClient.Machines(machine.Namespace).Update(machine)
+		return err
```
enxebre: Since we are dropping polling this won't reflect the real state of the instance in the cloud since we are returning true regardless the operation status.
Even if a new reconciliation is triggered by `a.machineClient.Machines(machine.Namespace).Update(machine)`  It's not guaranteed that then `Exist()`->`Update()` sets the real machine state/IPs.
In aws we validate the status returned by the api, that happens to be "pending" so we requeueAfter until the cloud state is "success", then we return true and the requeing loop is broken. We need to put a mechanism to emulate the behaviour above here.

frobware:
Need to decide where we want to go with this.

PR #19, opened 2019-05-20
Make create and delete async operations

frobware:
The previous behaviour for create was to block until the create operation was _"DONE"_.

This changes the overall behaviour of `create()` to be an asynchronous / idempotent call. If you pass a `UID` as an additional request parameter to `computeservice.InsertInstances()` then, as long as all the other parameters are equal (i.e., project, zone, name), this then becomes an idempotent call.

The create sequence is now:
- create instance from scope
- call `InstancesInsert` with instance
- check to see if operation is _"DONE"_
- requeue operation if not _"DONE"_
- when done, update spec
- when done, update status with created condition

The sentinel value for `exists()` now changes to:
- instance exists, and
- created condition exists on machine status

After a successful create, update will be called naturally - naturally because create would have modified the machine object, the `controller.Reconcile()` loop will trigger again (as part of the informer update on the machine object), and during this pass `exists()` will be `true`, so `update()` will be called.

Also:
- Updated the unit tests to reflect and handle the new async behaviour.
- The `MachineCreated` condition is now only set once create concludes

### DIFF
--- a/pkg/cloud/gcp/actuators/machine/reconciler.go
+++ b/pkg/cloud/gcp/actuators/machine/reconciler.go
@@ -15,17 +15,14 @@ import (
 	apicorev1 "k8s.io/api/core/v1"
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const (
-	operationDone      = "DONE"
-	operationRetryWait = 5 * time.Second
-	operationTimeOut   = 180 * time.Second
-	requeuePeriod      = 10 * time.Second
-	userDataSecretKey  = "userData"
+	operationDone     = "DONE"
+	requeuePeriod     = 10 * time.Second
+	userDataSecretKey = "userData"
 )
 
 // Reconciler are list of services required by machine actuator, easy to create a fake
@@ -218,29 +215,6 @@ func (r *Reconciler) getCustomUserData() (string, error) {
 		return "", fmt.Errorf("secret %v/%v does not have %q field set. Thus, no user data applied when creating an instance", r.machine.GetNamespace(), r.providerSpec.UserDataSecret.Name, userDataSecretKey)
 	}
 	return base64.StdEncoding.EncodeToString(data), nil
-}
-
-func (r *Reconciler) waitUntilOperationCompleted(zone, operationName string) (*compute.Operation, error) {
-	var op *compute.Operation
-	var err error
-	return op, wait.Poll(operationRetryWait, operationTimeOut, func() (bool, error) {
-		op, err = r.computeService.ZoneOperationsGet(r.projectID, zone, operationName)
-		if err != nil {
-			return false, err
-		}
-		klog.V(3).Infof("Waiting for %q operation to be completed... (status: %s)", op.OperationType, op.Status)
-		if op.Status == "DONE" {
-			if op.Error == nil {
-				return true, nil
-			}
-			var err []error
-			for _, opErr := range op.Error.Errors {
-				err = append(err, fmt.Errorf("%s", *opErr))
-			}
-			return false, fmt.Errorf("the following errors occurred: %+v", err)
-		}
-		return false, nil
-	})
 }
 
 func validateMachine(machine machinev1.Machine, providerSpec v1beta1.GCPMachineProviderSpec) error {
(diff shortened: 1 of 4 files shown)


pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -175,13 +207,6 @@ func (r *Reconciler) reconcileMachineWithCloudState(failedCondition *v1beta1.GCP
 		r.machine.Status.Addresses = nodeAddresses
 		r.providerStatus.InstanceState = &freshInstance.Status
 		r.providerStatus.InstanceID = &freshInstance.Name
-		succeedCondition := v1beta1.GCPMachineProviderCondition{
-			Type:    v1beta1.MachineCreated,
-			Reason:  machineCreationSucceedReason,
-			Message: machineCreationSucceedMessage,
-			Status:  corev1.ConditionTrue,
-		}
-		r.providerStatus.Conditions = reconcileProviderConditions(r.providerStatus.Conditions, succeedCondition)
```
michaelgugino: Are we sure we want to remove this?

frobware: It is set when Create is successful. How often can create successfully happen? It is also the new sentinel in exists - the cloud instance must exist and the machine itself must have the status condition. We could add an annotation, but that's just more of the same.

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -243,16 +268,20 @@ func (r *Reconciler) exists() (bool, error) {
...
 	}
-	_, err := r.computeService.InstancesGet(r.projectID, zone, r.machine.Name)
-	if err == nil {
-		klog.Infof("Machine %q already exists", r.machine.Name)
-		return true, nil
+
+	instance, err := r.computeService.InstancesGet(r.projectID, r.providerSpec.Zone, r.machine.Name)
+	if err != nil {
+		e, ok := err.(*googleapi.Error)
+		if ok && e.Code == 404 {
```
michaelgugino: We have a function for this.

frobware: I inlined it as it was only called once.

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -117,30 +119,60 @@ func (r *Reconciler) create() error {
...
-			klog.Errorf("Failed to reconcile machine with cloud state: %v", reconcileWithCloudError)
-		}
-		return fmt.Errorf("failed to wait for create operation via compute service. Operation status: %v. Error: %v", op, err)
+
+	// Passing machine.UID makes this request and subsequent
+	// requests idempotent as long as _all_ the other parameters
+	// to InstancesInsert() are equal. Note: you cannot use the
+	// same UID for any other operation as it is now keyed to the
+	// insert operation.
+	operation, err := r.computeService.InstancesInsert(string(r.machine.UID), r.projectID, r.providerSpec.Zone, instance)
```
michaelgugino: Just a follow-on from our discussion from earlier:  This would make the create event immutable in this actuator, which is currently different from what we do in AWS (aws has code to remove stopped instances, apparently).  I'm in favor of the immutable approach as done here.

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -117,30 +119,60 @@ func (r *Reconciler) create() error {
...
+	if err != nil {
+		return errors.Wrap(err, "InstancesInsert() failed")
 	}
-	return r.reconcileMachineWithCloudState(nil)
+
+	if operation.Status != operationDone {
+		klog.Infof("Create Operation #%v not %q for machine %q, returning an error to requeue", operation.Id, operationDone, r.machine.Name)
```
enxebre: when we not succeed we want to reflect it into the machine state/providerConditions  the likes of ```.reconcileMachineWithCloudState(&v1beta1.GCPMachineProviderCondition{		// Passing machine.UID makes this request and subsequent
			Type:    v1beta1.MachineCreated,		// requests idempotent as long as _all_ the other parameters
			Reason:  machineCreationFailedReason,		// to InstancesInsert() are equal. Note: you cannot use the
			Message: err.Error(),		// same UID for any other operation as it is now keyed to the
			Status:  corev1.ConditionFalse,		// insert operation.
		}); reconcileWithCloudError != nil {```
See a tested example of current approach https://github.com/openshift/cluster-api-provider-gcp/pull/14#issue-278378481

frobware: Doesn't this only apply to `InstancesInsert()`? Polling `operation.Status==DONE` will yield no error.

enxebre: why wouldn't we want to reflect this "progressing" InstanceState on the machine status?

frobware:
/hold 

Looking at another way

frobware:
Closing in favour of #22.

PR #21, opened 2019-05-20
Make delete non blocking

frobware:
The core machine controller will not call us back on error or an
explicit request to re-queue by returning RequeueAfterError.

By the time delete is called the machine object has gone. If were to
call machine.Update() and add an annotation for the operation.Id (so
that we could keep track of it in follow-up calls) the Update() will
fail.

Ideally we would like to return nil when the cloud operation is
"DONE". However, the current design of the core controller does not
allow for an idempotent delete operation.

### DIFF
--- a/pkg/cloud/gcp/actuators/machine/reconciler.go
+++ b/pkg/cloud/gcp/actuators/machine/reconciler.go
@@ -255,7 +255,7 @@ func (r *Reconciler) exists() (bool, error) {
 	return false, fmt.Errorf("error getting running instances: %v", err)
 }
 
-// Returns true if machine exists.
+// delete deletes the cloud instance, if and only if it exists
 func (r *Reconciler) delete() error {
 	exists, err := r.exists()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -265,14 +265,24 @@ func (r *Reconciler) delete() error {
...
+	// operation.
+	operation, err := r.computeService.InstancesDelete(r.projectID, r.providerSpec.Zone, r.machine.Name)
 	if err != nil {
-		return fmt.Errorf("failed to delete instance via compute service: %v", err)
-	}
-	if op, err := r.waitUntilOperationCompleted(zone, operation.Name); err != nil {
-		return fmt.Errorf("failed to wait for delete operation via compute service. Operation status: %v. Error: %v", op, err)
+		return fmt.Errorf("failed to delete instance for machine %q via compute service: %v", machineID(r.machine), err)
 	}
+	klog.Infof("Delete Operation #%v initiated for machine %q", operation.Id, machineID(r.machine))
```
michaelgugino: I'm unsure of how it might happen, but would the delete operation potentially not take effect for some reason?  If so, then we will have removed the machine-record, we'll have no indication later that an instance is left hanging around.  Maybe we should always fail here, and just let Exists() handle it in a later run.

frobware:
This looks to be self-inflicted. Will rework.

PR #22, opened 2019-05-22
actuator: make create and delete asynchronous operations

frobware:
The previous behaviour for `Reconciler.create()` and `Reconciler.delete()` was to block until their respective operations were complete. This change modifies both create() and delete() to be asynchronous operations. Each method returns a `RequeueRequestError()` until the underlying GCP cloud operations are marked as _DONE_.

### DIFF
--- a/pkg/cloud/gcp/actuators/machine/reconciler.go
+++ b/pkg/cloud/gcp/actuators/machine/reconciler.go
@@ -146,21 +146,19 @@ func (r *Reconciler) create() error {
 		}
 		operation, err = r.computeService.InstancesInsert(r.projectID, r.providerSpec.Zone, instance)
 		if err != nil {
-			if reconcileWithCloudError := r.reconcileMachineWithCloudState(&v1beta1.GCPMachineProviderCondition{
+			r.providerStatus.Conditions = setProviderConditions(r.providerStatus.Conditions, v1beta1.GCPMachineProviderCondition{
 				Type:    v1beta1.MachineCreated,
 				Reason:  machineCreationFailedReason,
 				Message: err.Error(),
 				Status:  corev1.ConditionFalse,
-			}); reconcileWithCloudError != nil {
-				klog.Errorf("Failed to reconcile machine with cloud state: %v", reconcileWithCloudError)
-			}
+			})
 			return fmt.Errorf("failed to create instance via compute service: %v", err)
 		}
 		r.machine.Annotations[pendingCreateKey] = fmt.Sprintf("%v", operation.Id)
 	}
 
 	klog.Infof("Instance create operation #%v status=%q for machine %q", operation.Id, operation.Status, r.machine.Name)
-	if err := r.reconcileMachineWithCloudState(nil); err != nil {
+	if err = r.reconcileMachineWithCloudState(); err != nil {
 		return err
 	}
 
@@ -170,12 +168,19 @@ func (r *Reconciler) create() error {
 	}
 
 	delete(r.machine.Annotations, pendingCreateKey)
+	r.providerStatus.Conditions = setProviderConditions(r.providerStatus.Conditions, v1beta1.GCPMachineProviderCondition{
+		Type:    v1beta1.MachineCreated,
+		Reason:  machineCreationSucceedReason,
+		Message: machineCreationSucceedMessage,
+		Status:  corev1.ConditionTrue,
+	})
+
 	_, err = r.persist()
 	return err
 }
 
 func (r *Reconciler) update() error {
-	if err := r.reconcileMachineWithCloudState(nil); err != nil {
+	if err := r.reconcileMachineWithCloudState(); err != nil {
 		return err
 	}
 	_, err := r.persist()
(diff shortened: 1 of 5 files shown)


pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -117,30 +114,63 @@ func (r *Reconciler) create() error {
 		Items: metadataItems,
 	}
 
-	operation, err := r.computeService.InstancesInsert(r.projectID, zone, instance)
+	return instance, nil
+}
+
+// Create creates machine if and only if machine exists, handled by cluster-api
```
michaelgugino: This comment is confusing and I'm not sure what it means.

frobware: Yep - that was there before though. :-) :-1:

frobware: Done in https://github.com/openshift/cluster-api-provider-gcp/commit/f3071fbeefc793d7b15ec1d45178e71138081ac7

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -155,47 +185,49 @@ func (r *Reconciler) reconcileMachineWithCloudState(failedCondition *v1beta1.GCP
...
-			Address: fmt.Sprintf("%s.c.%s.internal", r.machine.Name, r.projectID),
-		})
+	nodeAddresses := []v1.NodeAddress{{Type: v1.NodeInternalIP, Address: networkInterface.NetworkIP}}
+	for _, config := range networkInterface.AccessConfigs {
+		nodeAddresses = append(nodeAddresses, v1.NodeAddress{Type: v1.NodeExternalIP, Address: config.NatIP})
+	}
+	// Since we don't know when the project was created, we must account for
+	// both types of internal-dns:
+	// https://cloud.google.com/compute/docs/internal-dns#instance-fully-qualified-domain-names
+	// [INSTANCE_NAME].[ZONE].c.[PROJECT_ID].internal (newer)
```
bison: It looks like you can force the new style on a per-project, or per-instance basis:
https://cloud.google.com/compute/docs/internal-dns#setting-zonal-dns

Should we set `VmDnsSetting=ZonalOnly` and avoid this?

frobware: IIRC, didn't @michaelgugino mention that it depends on when your account was created. (I could have that recollection totally wrong.)

bison: The default depends on that, but I think you can override it.

pkg/cloud/gcp/actuators/machine/reconciler_test.go
```diff
@@ -4,27 +4,37 @@ import (
 	"fmt"
 	"testing"
 
+	"github.com/google/uuid"
```
michaelgugino: Dead code here?

bison: The import of the uuid package? It's used in creating the machine object below, right?

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -117,30 +114,63 @@ func (r *Reconciler) create() error {
...
+	if err := r.reconcileMachineWithCloudState(nil); err != nil {
+		return err
+	}
+
+	if operation.Status != "DONE" {
+		klog.Infof("Instance create operation #%v incomplete, returning an error to requeue", operation.Id)
+		return &clustererror.RequeueAfterError{RequeueAfter: requeuePeriod}
+	}
+
+	delete(r.machine.Annotations, pendingCreateKey)
```
michaelgugino: We should add a comment here that (hopefully) scope.close() will sync object state to make this remove effective.

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -247,45 +253,48 @@ func validateMachine(machine machinev1.Machine, providerSpec v1beta1.GCPMachineP
...
-	}
-	if isNotFoundError(err) {
-		klog.Infof("Machine %q does not exist", r.machine.Name)
-		return false, nil
+	instance, err := r.instanceGet()
+	if err != nil {
+		return false, err
 	}
-	return false, fmt.Errorf("error getting running instances: %v", err)
+	return instance != nil && !haveCreateOperationInProgress(r.machine), nil
```
michaelgugino: We can save api calls to the cloud provider by testing haveCreateOperationInProgress first.

```go
if haveCreateOperationInProgress(r.machine) {
  return false, nil
}
<get instance, check error>
return instance != nil
```

pkg/cloud/gcp/actuators/machine/reconciler_test.go
michaelgugino: @bison yes, but we're not consuming that data anywhere anymore.  We don't need to mock that up for the tests.

pkg/cloud/gcp/actuators/machine/reconciler.go
michaelgugino: Actually, this would save 2 api calls per loop if we're not finished creating.

pkg/cloud/gcp/actuators/machine/reconciler.go
michaelgugino: Sure, but we're not in charge of creating a user's project, that is same as an aws account.  Enforcing a project-wide setting is probably not an option for us.

pkg/cloud/gcp/actuators/machine/reconciler_test.go
frobware: This has all gone - I think I borked because I needed to rebase.

pkg/cloud/gcp/actuators/machine/reconciler.go
frobware: As things stand today I wanted to the double check, even if it is more costly.

If the annotation is left on the machine (because scope.Close() failed) then this would always return true - which is only one half of the story.

pkg/cloud/gcp/actuators/machine/reconciler.go
frobware: The whole `scope.Close()` pattern seems so fragile to me.

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -117,30 +114,63 @@ func (r *Reconciler) create() error {
...
+				klog.Errorf("Failed to reconcile machine with cloud state: %v", reconcileWithCloudError)
+			}
+			return fmt.Errorf("failed to create instance via compute service: %v", err)
+		}
+		r.machine.Annotations[pendingCreateKey] = fmt.Sprintf("%v", operation.Id)
 	}
-	return r.reconcileMachineWithCloudState(nil)
+
+	klog.Infof("Instance create operation #%v status=%q for machine %q", operation.Id, operation.Status, r.machine.Name)
+	if err := r.reconcileMachineWithCloudState(nil); err != nil {
```
enxebre: we need to make the harcoded `succeedCondition` set by `reconcileMachineWithCloudState`  a parameter or decouple it

pkg/cloud/gcp/actuators/machine/reconciler.go
frobware: > We should add a comment here that (hopefully) scope.close() will sync object state to make this remove effective.

I moved all `scope.Close()` to their respective functions in the reconciler.

pkg/cloud/gcp/actuators/machine/reconciler.go
frobware: Done in 96561e7

frobware:
Closing in favour of https://github.com/openshift/cluster-api-provider-gcp/pull/23.

Will do delete as a separate PR.

PR #35, opened 2019-07-23
Do not b64-encode user data

csrwng:
The user data that is passed to the GCP api does not need to be base64-encoded


csrwng:
@ingvagabund ptal

PR #36, opened 2019-07-23
Allow specifying whether you want a public IP

csrwng:
Currently a public IP is generated for every network interface in a machine (by specifying a non-empty AccessConfig). This PR introduces a boolean to allow you to specify whether a public IP is needed. For OCP use cases it's not.

### DIFF
--- a/pkg/cloud/gcp/actuators/machine/reconciler.go
+++ b/pkg/cloud/gcp/actuators/machine/reconciler.go
@@ -71,9 +71,14 @@ func (r *Reconciler) create() error {
 
 	// networking
 	var networkInterfaces = []*compute.NetworkInterface{}
+
 	for _, nic := range r.providerSpec.NetworkInterfaces {
+		accessConfigs := []*compute.AccessConfig{}
+		if nic.PublicIP {
+			accessConfigs = append(accessConfigs, &compute.AccessConfig{})
+		}
 		computeNIC := &compute.NetworkInterface{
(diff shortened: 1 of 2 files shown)


csrwng:
@ingvagabund ptal

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -71,9 +71,14 @@ func (r *Reconciler) create() error {
...
 	// networking
 	var networkInterfaces = []*compute.NetworkInterface{}
+
 	for _, nic := range r.providerSpec.NetworkInterfaces {
+		accessConfigs := []*compute.AccessConfig{}
+		if nic.PublicIP {
+			accessConfigs = append(accessConfigs, &compute.AccessConfig{})
+		}
 		computeNIC := &compute.NetworkInterface{
-			AccessConfigs: []*compute.AccessConfig{{}},
```
ingvagabund: I will not ask how this is working at the first place. I.e. empty `&compute.AccessConfig{}` item without any additional fields set. Though, +1 for making publicIP optional.

csrwng: Yeah, it just needs to be present for you to get a public IP (works same way in Terraform).

ingvagabund: Strange world we live in :) Thanks for the explanation.

PR #37, opened 2019-07-24
Pass full URL to image when creating compute instance

csrwng:
The image parameter must be in the form:
https://www.googleapis.com/compute/v1/projects/{project}/global/images/{image}

### DIFF
--- a/pkg/cloud/gcp/actuators/machine/reconciler.go
+++ b/pkg/cloud/gcp/actuators/machine/reconciler.go
@@ -63,7 +63,7 @@ func (r *Reconciler) create() error {
 				DiskSizeGb:  disk.SizeGb,
 				DiskType:    fmt.Sprintf("zones/%s/diskTypes/%s", zone, disk.Type),
 				Labels:      disk.Labels,
-				SourceImage: disk.Image,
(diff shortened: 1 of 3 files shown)


pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -63,7 +63,7 @@ func (r *Reconciler) create() error {
 				DiskSizeGb:  disk.SizeGb,
 				DiskType:    fmt.Sprintf("zones/%s/diskTypes/%s", zone, disk.Type),
 				Labels:      disk.Labels,
-				SourceImage: disk.Image,
+				SourceImage: fmt.Sprintf("%s%s/global/images/%s", r.computeService.BasePath(), r.projectID, disk.Image),
```
ingvagabund: Can we assume `r.computeService.BasePath()` will always end with `/`? Will the source image path be still valid even with `%s/%s/global/images/%s`?

csrwng: It does from what I've seen, but we can use the googleapi.ResolveRelative function to be on the safe side, will update.

csrwng:
Updated

ingvagabund:
Awesome!!! thanks

/approve
/lgtm

PR #38, opened 2019-07-24
Fix linking on other architectures

yselkowitz:
This avoids actual static linking of -lpthread,-lc,etc. on some architectures.


mkumatag:
@yselkowitz any idea why this PR been closed?

yselkowitz:
@mkumatag I was informed that GCP's Power support is limited to AIX.

mkumatag:
> @mkumatag I was informed that GCP's Power support is limited to AIX.

well, that's true. Will revisit if we need it in the future.

PR #40, opened 2019-07-26
Add support for TargetPools

michaelgugino:
This commit extends the providerSpec to add the field
TargetPools, which is a list of strings specifying
targetpools you wish to add the instance to.

This commit supports removing the instance from a
target pool when the machine-object is deleted.

This commit also adds unit tests.

The target pools must already exist for this to work, similiar to AWS LB's.
 If we don't have this PR, masters won't be added to the appropriate LB's when doing disaster recovery.

### DIFF
--- a/pkg/cloud/gcp/actuators/machine/reconciler.go
+++ b/pkg/cloud/gcp/actuators/machine/reconciler.go
@@ -20,6 +20,7 @@ import (
 const (
 	userDataSecretKey   = "userData"
 	requeueAfterSeconds = 20
+	instanceLinkFmt     = "https://www.googleapis.com/compute/v1/projects/%s/zones/%s/instances/%s"
 )
 
 // Reconciler are list of services required by machine actuator, easy to create a fake
@@ -136,6 +137,10 @@ func (r *Reconciler) create() error {
 }
 
 func (r *Reconciler) update() error {
+	// Add target pools, if necessary
+	if err := r.processTargetPools(true, r.addInstanceToTargetPool); err != nil {
+		return err
+	}
 	return r.reconcileMachineWithCloudState(nil)
 }
 
@@ -217,6 +222,13 @@ func validateMachine(machine machinev1.Machine, providerSpec v1beta1.GCPMachineP
 	// TODO (alberto): First validation should happen via webhook before the object is persisted.
 	// This is a complementary validation to fail early in case of lacking proper webhook validation.
 	// Default values can also be set here
+	if providerSpec.TargetPools != nil {
+		for _, pool := range providerSpec.TargetPools {
+			if pool == "" {
+				return fmt.Errorf("all target pools must have valid name")
+			}
+		}
+	}
 	return nil
 }
 
@@ -251,6 +263,10 @@ func (r *Reconciler) exists() (bool, error) {
 
 // Returns true if machine exists.
 func (r *Reconciler) delete() error {
+	// Remove instance from target pools, if necessary
+	if err := r.processTargetPools(false, r.deleteInstanceFromTargetPool); err != nil {
+		return err
+	}
 	exists, err := r.exists()
 	if err != nil {
 		return err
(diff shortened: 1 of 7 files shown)


pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -135,7 +136,16 @@ func (r *Reconciler) create() error {
 	return r.reconcileMachineWithCloudState(nil)
 }
 
+func fmtInstanceSelfLink(project, zone, name string) string {
+	return fmt.Sprintf(instanceLinkFmt, project, zone, name)
+}
+
 func (r *Reconciler) update() error {
+	// Add target pools, if necessary
+	err := r.processTargetPools(r.addInstanceToTargetPool)
```
enxebre: inline?

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -278,3 +300,73 @@ func isNotFoundError(err error) bool {
...
+			if err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func (r *Reconciler) addInstanceToTargetPool(instanceLink string, pool string) error {
+	inPool, err := r.instanceExistsInPool(instanceLink, pool)
```
enxebre: what happens if you don't check and just add?

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -132,9 +135,66 @@ func (r *Reconciler) create() error {
...
+	instanceInfo := struct {
+		Project string
+		Zone    string
+		Name    string
+	}{
+		Project: project,
+		Zone:    zone,
+		Name:    name,
+	}
+	tmpl, err := template.New("test").Parse(instanceLinkTString)
```
frobware: Why isn't this just sprintf? That would get rid of the panic too.

pkg/cloud/gcp/actuators/machine/reconciler.go
michaelgugino: nothing.  If you just add repeatedly, it will just succeed.  Technically speaking, we might not need to check if it exists, but I think having the check might make feedback to the user more consumable if something fails; eg, did the get fail, or the add-instance fail (Did user specify a target pool that doesn't exist)?  Ideally, we only do this once during create, and for all but the initial time this runs, we'll only be doing 1 api call: 2 calls first time, get and add, every other time, just get.

pkg/cloud/gcp/actuators/machine/reconciler.go
michaelgugino: yeah, I didn't realize it was going to be so verbose to write it this way, but I did, lol.  I already planned on replacing it.

michaelgugino:
Tested this on a live cluster, everything seems to work well and the logs are clean: https://gist.github.com/michaelgugino/018fa22dd26251e49d06125b54bdeb07

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -278,3 +294,66 @@ func isNotFoundError(err error) bool {
...
+			}
+		}
+	}
+	return false, nil
+}
+
+type poolProcessor func(instanceLink, pool string) error
+
+func (r *Reconciler) processTargetPools(desired bool, poolFunc poolProcessor) error {
+	if r.providerSpec.TargetPools != nil && len(r.providerSpec.TargetPools) > 0 {
```
frobware: nit; return early if this condition does not hold true. will help reduce the indentation.

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -278,3 +294,68 @@ func isNotFoundError(err error) bool {
...
+			}
+		}
+	}
+	return false, nil
+}
+
+type poolProcessor func(instanceLink, pool string) error
+
+func (r *Reconciler) processTargetPools(desired bool, poolFunc poolProcessor) error {
+	if r.providerSpec.TargetPools == nil && len(r.providerSpec.TargetPools) == 0 {
```
frobware: Shouldn't this condition be _OR_?

michaelgugino: yes, it should be!

enxebre: can we please cover this in the unit?

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -278,3 +294,68 @@ func isNotFoundError(err error) bool {
...
+			}
+		}
+	}
+	return false, nil
+}
+
+type poolProcessor func(instanceLink, pool string) error
+
+func (r *Reconciler) processTargetPools(desired bool, poolFunc poolProcessor) error {
+	if len(r.providerSpec.TargetPools) == 0 {
```
frobware: You can drop this condition; it is safe to iterate over a nil slice.

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -278,3 +294,68 @@ func isNotFoundError(err error) bool {
...
+
+func (r *Reconciler) instanceExistsInPool(instanceLink string, pool string) (bool, error) {
+	// Get target pool
+	tp, err := r.computeService.TargetPoolsGet(r.projectID, r.providerSpec.Region, pool)
+	if err != nil {
+		return false, fmt.Errorf("unable to get targetpool: %v", err)
+	}
+
+	// Check if instance is in target pool already
+	if tp.Instances != nil {
```
frobware: You can drop this condition; it is safe to iterate over a nil slice.

pkg/cloud/gcp/actuators/machine/reconciler.go
michaelgugino: That's true, but saving the call to generate instanceSelfLink is string manipulation and is expensive compared to this check.

frobware: Move that to where the condition holds true.

frobware:
"This commit adds unit tests" -  should there be unit test that cover:
- `addInstanceToTargetPool()`
- `deleteInstanceFromTargetPool()`

because running cover[1] I don't see these functions being hit at.

[1] - cover is dumb utility I use for checking coverage as I'm writing tests.

```bash
$ type cover
cover is a function
cover () 
{ 
    local t="/tmp/go-cover.$$.tmp";
    go test -coverprofile=$t $@ && go tool cover -html=$t && unlink $t
}
```

pkg/cloud/gcp/actuators/machine/reconciler.go
frobware: > That's true, but saving the call to generate instanceSelfLink is string manipulation and is expensive compared to this check.

OK; but given the amount of general logging we do I think calling fmt here for the nil cases... is no worse than just another log statement.

michaelgugino:
> "This commit adds unit tests" - should there be unit test that cover:
> 
>     * `addInstanceToTargetPool()`
> 
>     * `deleteInstanceFromTargetPool()`
> 
> 
> because running cover[1] I don't see these functions being hit.
> 
> [1] - cover is dumb utility I use for checking coverage as I'm writing tests.
> 
> ```shell
> $ type cover
> cover is a function
> cover () 
> { 
>     local t="/tmp/go-cover.$$.tmp";
>     go test -coverprofile=$t $@ && go tool cover -html=$t && unlink $t
> }
> ```

@frobware yeah, I don't see a lot of value in adding tests for those functions as they'll be covered by integration tests.  It would be like 'does this function call this other mocked function and return either an error or no error' is not very useful test.

PR #41, opened 2019-07-30
Introduce gosec for security checks

JAORMX:
gosec does static code analysis and checks for common security issues in
golang codebases.

This PR introduces the tool, and sets it up as a test target in the Makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,6 +34,10 @@ vet: ## Apply go vet to all go files
 unit: # Run unit test
 	$(DOCKER_CMD) go test -race -cover ./cmd/... ./pkg/...
 
+.PHONY: sec
+sec: # Run security static analysis
+	hack/gosec.sh ./...
+
 .PHONY: build
 build: ## build binaries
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -30,6 +30,11 @@ vet: ## Apply go vet to all go files
 unit: # Run unit test
 	$(DOCKER_CMD) go test -race -cover ./cmd/... ./pkg/...
 
+.PHONY: sec
+sec: # Run security static analysis
+	go get -u github.com/securego/gosec/cmd/gosec
+	gosec -severity medium -confidence medium -exclude G304 -quiet ./...
```
ingvagabund: Would it be possible to run the command from within a container?

JAORMX: It is possible. That's actually what was proposed for the installer: https://github.com/openshift/installer/pull/2126/files

JAORMX:
Any more feedback on this PR?

Makefile
ingvagabund: Can you update the PR so `gosec` is run inside a container?

JAORMX: Done! thanks for the feedback

PR #43, opened 2019-08-01
Add project ID to machine provider spec

csrwng:
The project associated with a particular GCP credential key may not be the same project that was specified for the cluster install. The service account associated with the credential key may exist in one project and manage a cluster in a different project. Adding a project ID to the machine provider spec ensures that the machine controller uses the intended project in the edge case where it's different than the project associated with the credential key.

If projectID is not specified on machine, use the one from the credential secret.


enxebre:
needs to regenerate `zz_generated.deepcopy.go`

PR #51, opened 2019-08-27
bug 1741765: Allow invalid project/zone machines to be deleted

michaelgugino:
Currently, if someone creates a machine with an invalid
project or zone, a machine will never be created.  Validation
and code reuse prevents proper deletion.

Ignore and not checking project/zone is not an option because
GCP returns a 404 error just as if a machine was not found
in a valid project.

This commit allows deletion of an invalid machine that does
not have status.addresses.  This ensures that the machine
actually was not created, and that the object was not
later made invalid by the user.


michaelgugino:
I tested this patch on a live cluster, it works as expected.

ingvagabund:
/retitle bug 1741765: Allow invalid project/zone machines to be deleted

PR #54, opened 2019-08-29
Update machine only when changed

ingvagabund:


Gopkg.toml
```diff
@@ -58,9 +58,8 @@ required = [
 
 [[override]]
   name = "sigs.k8s.io/controller-runtime"
-  # Pin the latest master HEAD to get support webhooks
-  # which are not yet released
-  revision="96b67f231945a8e8550dbdd8bfbd550a0c68097b"
+  # release-0.2 since it imports kubernetes 1.14
+  branch="release-0.2"
```
frobware: Nice!

pkg/cloud/gcp/actuators/machine/actuator.go
```diff
@@ -70,6 +70,8 @@ func (a *Actuator) Create(ctx context.Context, cluster *clusterv1.Cluster, machi
 		return a.handleMachineError(machine, fmt.Errorf(fmtErr), createEventAction)
 	}
 	if err := newReconciler(scope).create(); err != nil {
+		// Update machine and machine status in case it was modified
+		scope.Close()
```
frobware: This is a nit; if err is a `ConflictError` then we could skip calling `scope.Close()`.

PR #55, opened 2019-08-30
WIP: Drain unready

michaelgugino:


vendor/github.com/openshift/kubernetes-drain/drain.go
```diff
@@ -506,12 +508,22 @@ func waitForDelete(pods []corev1.Pod, interval, timeout time.Duration, usingEvic
 				logf(logger, "pod %q removed (%s)", pod.Name, verbStr)
```
michaelgugino: For some unknown reason, this patchset makes the code enter this path when it definitely shouldn't.  It's just plan crazy.

PR #57, opened 2019-08-30
Label instances created by the actuator with cluster ID

ingvagabund:
So the installer can potentially delete all gcp instances when a cluster is destroyed.

Label key constraints: The key must start with a lowercase character,
can only contain lowercase letters, numeric characters, underscores and dashes.
The key can be at most 63 characters long. International characters are allowed.

Thus, using "kubernetes-io-cluster-CLUSTERID=owned" form instead.

```
$ gcloud compute instances list --format='table(name,status,labels.list())' | grep jchalo-cf757
jchalo-cf757-m-0                        RUNNING
jchalo-cf757-w-a-jwddf                  RUNNING
jchalo-cf757-m-1                        RUNNING
jchalo-cf757-w-b-92xt6                  RUNNING  kubernetes-io-cluster-jchalo-cf757=owned
jchalo-cf757-w-b-wd8nm                  RUNNING
jchalo-cf757-m-2                        RUNNING
jchalo-cf757-w-c-54v9j                  RUNNING
```

### DIFF
--- a/pkg/cloud/gcp/actuators/machine/reconciler.go
+++ b/pkg/cloud/gcp/actuators/machine/reconciler.go
@@ -54,6 +54,11 @@ func (r *Reconciler) create() error {
 		},
 	}
 
+	if instance.Labels == nil {
+		instance.Labels = map[string]string{}
+	}
+	instance.Labels[fmt.Sprintf("kubernetes-io-cluster-%v", r.machine.Labels[machinev1.MachineClusterIDLabel])] = "owned"
+
 	// disks
 	var disks = []*compute.AttachedDisk{}
 	for _, disk := range r.providerSpec.Disks {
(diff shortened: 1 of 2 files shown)


pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -42,11 +42,18 @@ func (r *Reconciler) create() error {
 		return fmt.Errorf("failed validating machine provider spec: %v", err)
 	}
 
+	instanceLabels := map[string]string{}
+	for key, value := range r.providerSpec.Labels {
+		instanceLabels[key] = value
+	}
+
+	instanceLabels[fmt.Sprintf("kubernetes-io-cluster-%v", r.machine.Labels[machinev1.MachineClusterIDLabel])] = "owned"
```
frobware: Any reason why don't avoid the copy and just do this after we construct:

```go
instance := &compute.Instance{...}
instance.Labels[fmt.Sprintf("kubernetes-io-cluster-%v", r.machine.Labels[machinev1.MachineClusterIDLabel])] = "owned"

ingvagabund: Looks better!!! Done.

frobware:
So I read https://bugzilla.redhat.com/show_bug.cgi?id=1742227 and it's not clear whether this will be helpful. So cancelling LGTM for the moment.

/lgtm cancel

frobware:
From @abhinavdahiya: "The tags in gcp have meaning in terms of networking. Don't add tags.
Labels sure."

/lgtm

PR #59, opened 2019-09-03
WIP: Allow drain to continue when node unready

michaelgugino:
Sometimes, nodes go unready.  If we're deleting a machine,
this might even be expected.  When attempting to drain
an unready node, if that node has any pods with
'local storage', those pods will never be successfully
removed since the node cannot confirm their removal.

This commit allows us to ignore pods with a deletion
timestamp older than 5 minutes when a node is unready.
This will allow us to remove problematic nodes without
disrupting PDBs or skipping appropriate drain steps.

### DIFF
--- a/vendor/github.com/openshift/cluster-api/pkg/drain/drain.go
+++ b/vendor/github.com/openshift/cluster-api/pkg/drain/drain.go
@@ -23,7 +23,7 @@ import (
 	"math"
 	"sort"
 	"strings"
-	"sync"
+	//"sync"
 	"time"
 
 	golog "github.com/go-log/log"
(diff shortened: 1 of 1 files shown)


vendor/github.com/openshift/cluster-api/pkg/drain/drain.go
```diff
@@ -334,6 +334,12 @@ func getPodsForDeletion(client kubernetes.Interface, node *corev1.Node, options
 	}
 
 	for _, pod := range podList.Items {
+		if !pod.ObjectMeta.DeletionTimestamp.IsZero() && time.Now().Sub(pod.ObjectMeta.GetDeletionTimestamp().Time).Minutes() > 1 {
```
michaelgugino: This seems to work, but for some reason, we appear to be processing the deletes serially.

michaelgugino: https://gist.github.com/michaelgugino/ebb5b60f6e1193d93b5f55a96e5c92e5

enxebre:
This is meant for when node is unreachable. Covered by https://github.com/openshift/cluster-api-provider-gcp/pull/74
/close

PR #60, opened 2019-09-04
Fix linking on other architectures

yselkowitz:
This avoids actual static linking of -lpthread,-lc,etc. on some architectures.

This reopens https://github.com/openshift/cluster-api-provider-gcp/pull/38.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ WORKDIR /go/src/github.com/openshift/cluster-api-provider-gcp
 COPY . .
 # VERSION env gets set in the openshift/release image and refers to the golang version, which interfers with our own
 RUN unset VERSION \
- && GOPROXY=off NO_DOCKER=1 make build
+ && make build GOPROXY=off NO_DOCKER=1 CGO_ENABLED=0
 
 FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
 COPY --from=builder /go/src/github.com/openshift/cluster-api-provider-gcp/bin/manager /


Dockerfile
```diff
@@ -3,7 +3,7 @@ WORKDIR /go/src/github.com/openshift/cluster-api-provider-gcp
 COPY . .
 # VERSION env gets set in the openshift/release image and refers to the golang version, which interfers with our own
 RUN unset VERSION \
- && NO_DOCKER=1 make build
+ && make build NO_DOCKER=1 CGO_ENABLED=0
```
ingvagabund: Makefile:11-19
```
NO_DOCKER ?= 0
ifeq ($(NO_DOCKER), 1)
  DOCKER_CMD =
  IMAGE_BUILD_CMD = imagebuilder
  CGO_ENABLED = 1
else
  DOCKER_CMD := docker run --rm -e CGO_ENABLED=1 -v "$(PWD)":/go/src/github.com/openshift/cluster-api-provider-gcp:Z -w /go/src/openshift/cluster-api-provider-gcp openshift/origin-release:golang-1.12
  IMAGE_BUILD_CMD = docker build
endif
```

I don't think we use any C libraries directly in the code. So we might just set CGO_ENABLED to 0 by default.

frobware: Which other archs are we trying to support?

frobware: > I don't think we use any C libraries directly in the code. So we might just set CGO_ENABLED to 0 by default.

The `CGO_ENABLED = ` is unconditional. Change it to `CGO_ENABLED ?= 1`.

yselkowitz: > Which other archs are we trying to support?

IBM Power LE, for [IBM Power Systems for Google Cloud](https://console.cloud.google.com/marketplace/details/ibm-sg/ibm-power-cloud-for-gcp).

yselkowitz: > The `CGO_ENABLED = ` is unconditional. Change it to `CGO_ENABLED ?= 1`.

The difference in syntax only changes how the variable can be overridden.  `?=` allows also for environment variable overrides (e.g. `NO_DOCKER=1 make`), `=` allows only for argument overrides (e.g. `make CGO_ENABLED=0`).

yselkowitz:
The goimports test failure should be unrelated to this PR, could it please be /skip'ped?

yselkowitz:
Is anything else needed from me to get this merged?

yselkowitz:
Isn't 4.3 open yet?

PR #63, opened 2019-09-26
Error reporting

enxebre:
This is to satisfy contract with the machine controller so it can react appropriately and signal a machine as failed with the given error message when an InvalidMachineConfiguration error is reported, main cases are:
- The spec is invalid:
	Unable to be unmarshalled
	Missing expected labels by design
	Referencing non existing kubernetes resources
- The input values provoke a client error 4xx response from the cloud call to create a machine.

 openshift/enhancements#11
openshift/cluster-api#75


enxebre:
/hold
https://github.com/openshift/cluster-api/pull/75

PR #68, opened 2019-12-11
Add more unit tests for reconciler

alexander-demichev:


pkg/cloud/gcp/actuators/machine/reconciler_test.go
```diff
@@ -16,37 +19,154 @@ import (
...
+				Reason:  machineCreationFailedReason,
+				Message: "fail",
+			},
+			mockInstancesInsert: func(project string, zone string, instance *compute.Instance) (*compute.Operation, error) {
+				return nil, errors.New("fail")
+			},
+		},
+		{
+			name:          "Fail on google api error",
+			expectedError: errors.New("error launching instance: googleapi: Error 400: error"),
```
michaelgugino: We are returning a specific type of error in the function tested.  We should use that type.

enxebre:
thanks a bunch @alexander-demichev 
/lgtm
/retest

PR #70, opened 2020-01-06
Remove unnecessary machine specific client

alexander-demichev:
This PR removes unnecessary machine specific client, it can be replaced with controller runtime client.


enxebre:
can we please have a PR description with the motivation of this PR?

This PR is just removing the object at instantiation time, but it's still being passed around and relied upon.
https://github.com/openshift/cluster-api-provider-gcp/blob/master/pkg/cloud/gcp/actuators/machine/actuator.go#L35

https://github.com/openshift/cluster-api-provider-gcp/blob/master/pkg/cloud/gcp/actuators/machine/actuator.go#L43

https://github.com/openshift/cluster-api-provider-gcp/blob/master/pkg/cloud/gcp/actuators/machine/machine_scope.go#L149

enxebre:
/approve
@bison PTAL
@alexander-demichev can you please put a follow up to consolidate with this approach https://github.com/openshift/machine-api-operator/blob/master/pkg/controller/vsphere/machine_scope.go#L88 and use patch requests instead?

pkg/cloud/gcp/actuators/machine/machine_scope.go
```diff
@@ -146,11 +137,12 @@ func (s *machineScope) storeMachineSpec() error {
 
 	klog.V(4).Infof("Storing machine spec for %q, resourceVersion: %v, generation: %v", s.machine.Name, s.machine.ResourceVersion, s.machine.Generation)
 	s.machine.Spec.ProviderSpec.Value = ext
-	latestMachine, err := s.machineClient.Update(s.machine)
+
+	err = s.coreClient.Update(context.Background(), s.machine)
```
frobware: The remainder can be reduced to just:

```go
return s.coreClient.Update(context.Background(), s.machine)
```

pkg/cloud/gcp/actuators/machine/machine_scope.go
```diff
@@ -169,11 +161,10 @@ func (s *machineScope) storeMachineStatus() error {
 	s.machine.Status.ProviderStatus = ext
 	time := metav1.Now()
 	s.machine.Status.LastUpdated = &time
-	latestMachine, err := s.machineClient.UpdateStatus(s.machine)
+	err = s.coreClient.Status().Update(context.Background(), s.machine)
```
frobware: The remainder can be reduced to just:

```go
return s.coreClient.Status().Update(context.Background(), s.machine)
```

pkg/cloud/gcp/actuators/machine/machine_scope.go
```diff
@@ -169,11 +163,29 @@ func (s *machineScope) storeMachineStatus() error {
...
 	time := metav1.Now()
 	s.machine.Status.LastUpdated = &time
-	latestMachine, err := s.machineClient.UpdateStatus(s.machine)
-	if err != nil {
+
+	return nil
+}
+
+func (s *machineScope) PatchMachine() error {
+	klog.V(3).Infof("%v: patching machine", s.machine.GetName())
```
JoelSpeed: Nit: should this be a string instead?
```suggestion
	klog.V(3).Infof("%s: patching machine", s.machine.GetName())
```

PR #73, opened 2020-01-30
Add gitignore

alexander-demichev:
This PR adds .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,15 @@
+# Binaries for programs and plugins
+*.exe
+*.exe~
+*.dll
+*.so
+*.dylib
+
+# Test binary, build with `go test -c`
+*.test
+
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -0,0 +1,29 @@
...
+*.so
+*.dylib
+
+# Test binary, build with `go test -c`
+*.test
+
+# Output of the go coverage tool, specifically when used with LiteIDE
+*.out
+
+### Terraform ###
```
enxebre: curious why considering any terraform related?

alexander-demichev: No reason actually, I copied this file from AWS https://github.com/openshift/cluster-api-provider-aws/blob/master/.gitignore

enxebre: We use to run terraform for some testing there. I don't think we need this now.

PR #77, opened 2020-03-13
[OCPCLOUD-779] Add Machineset annotations to support autoscaling to/from zero

JoelSpeed:
This PR adds a new controller to annotate machinesets with information required to enable autoscaling to/from zero based on providerSpec input based on https://github.com/openshift/enhancements/pull/186.

Analogous to https://github.com/openshift/cluster-api-provider-aws/pull/301 and https://github.com/openshift/cluster-api-provider-azure/pull/112


JoelSpeed:
Have verified this works on a real cluster

enxebre:
Can we please include some links to the enhancement proposal and counterpart PRs as per https://github.com/openshift/cluster-api-provider-aws/pull/301#issue-386617237 in this PR desc and possibly the commit adding the controller?

pkg/cloud/gcp/actuators/machineset/cache.go
```diff
@@ -0,0 +1,60 @@
...
+	machinesCache map[MachineTypeKey]*gce.MachineType
+}
+
+// newMachineCache creates empty machineCache.
+func newMachineCache() *machineCache {
+	return &machineCache{
+		machinesCache: map[MachineTypeKey]*gce.MachineType{},
+	}
+}
+
```
enxebre: nit: comment seems like `GetMachineFromCache` is exported but it is not.

pkg/cloud/gcp/actuators/machineset/cache.go
```diff
@@ -0,0 +1,60 @@
...
+// GetMachineFromCache retrieves machine type from cache under lock.
+func (mc *machineCache) getMachineFromCache(gcpService computeservice.GCPComputeService, projectID string, zone string, machineType string) (*gce.MachineType, error) {
+	mc.cacheMutex.Lock()
+	defer mc.cacheMutex.Unlock()
+
+	// Machine Type already fetched from GCE
+	if machine, ok := mc.machinesCache[MachineTypeKey{zone, machineType}]; ok {
```
enxebre: s/machine/machineType in 49, 53 and 59?

pkg/cloud/gcp/actuators/machineset/cache.go
```diff
@@ -0,0 +1,60 @@
...
+
+// newMachineCache creates empty machineCache.
+func newMachineCache() *machineCache {
+	return &machineCache{
+		machinesCache: map[MachineTypeKey]*gce.MachineType{},
+	}
+}
+
+// GetMachineFromCache retrieves machine type from cache under lock.
+func (mc *machineCache) getMachineFromCache(gcpService computeservice.GCPComputeService, projectID string, zone string, machineType string) (*gce.MachineType, error) {
```
enxebre: `getMachineTypeFromCache`?

pkg/cloud/gcp/actuators/machineset/cache.go
```diff
@@ -0,0 +1,60 @@
...
+// MachineTypeKey is used to identify MachineType.
+type MachineTypeKey struct {
+	Zone        string
+	MachineType string
+}
+
+// machineCache is used for caching machine types.
+type machineCache struct {
+	cacheMutex    sync.Mutex
+	machinesCache map[MachineTypeKey]*gce.MachineType
```
enxebre: is this actually a `machineTypesCache`?

pkg/cloud/gcp/actuators/machineset/cache.go
JoelSpeed: One of the function args is `machineType` so I renamed it to `mt`, could also rename the function argument to something like `machineTypeName` but it's plumbed through from a field that is `MachineType`

pkg/cloud/gcp/actuators/machineset/cache.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+}
+
+// getMachineTypeFromCache retrieves machine type from cache under lock.
+func (mc *machineTypesCache) getMachineTypeFromCache(gcpService computeservice.GCPComputeService, projectID string, zone string, machineType string) (*gce.MachineType, error) {
+	mc.cacheMutex.Lock()
+	defer mc.cacheMutex.Unlock()
+
+	// Machine Type already fetched from GCE
+	if machine, ok := mc.machineTypesCache[machineTypeKey{zone, machineType}]; ok {
```
enxebre: nit: `if mt, ok... return mt`

enxebre:
@JoelSpeed this repo was actually never a fork so no need for `UPSTREAM: <carry>: ...` format.

pkg/cloud/gcp/actuators/machineset/cache.go
JoelSpeed: Fixed

JoelSpeed:
> @JoelSpeed this repo was actually never a fork so no need for `UPSTREAM: <carry>: ...` format.

I was just copying what I'd done on the Azure one tbh 😅

pkg/cloud/gcp/actuators/machineset/controller.go
```diff
@@ -0,0 +1,174 @@
...
+		return ctrl.Result{}, mapierrors.InvalidMachineConfiguration("error fetching machine type %q: %v", providerConfig.MachineType, err)
+	}
+
+	if machineSet.Annotations == nil {
+		machineSet.Annotations = make(map[string]string)
+	}
+
+	// TODO: get annotations keys from machine API
+	machineSet.Annotations[cpuKey] = strconv.FormatInt(machineType.GuestCpus, 10)
+	machineSet.Annotations[memoryKey] = strconv.FormatInt(machineType.MemoryMb, 10)
```
elmiko: do we need to add something here for the gpu value?

JoelSpeed: We don't have GPU implemented now so my thought was to leave it empty and allow the autoscaler to do the defaulting when it doesn't find the key.

Then when we implement GPU support we can add the correct value here and everything should carry on working.

I don't think we should be setting it to 0 explicitly here in case users manually add them, until we have a way to support, them omitting makes more sense to me

elmiko: sounds good to me!

> I don't think we should be setting it to 0 explicitly here in case users manually add them, until we have a way to support, them omitting makes more sense to me

agreed

PR #85, opened 2020-04-23
[OCPCLOUD-492] Implement leader election for gcp provider

Danil-Grigorev:
Couple of new cli arguments for configuring leader election:

```
  -leader-elect
    	Start a leader election client and gain leadership before executing the main loop. Enable this when running replicated components for high availability. (default true)
  -leader-elect-lease-duration int
    	The duration that non-leader candidates will wait after observing a leadership renewal until attempting to acquire leadership of a led but unrenewed leader slot. This is effectively the maximum duration that a leader can be stopped before it is replaced by another candidate. This is only applicable if leader election is enabled. (default 15s)
  -leader-elect-resource-namespace string
    	The namespace of resource object that is used for locking during leader election. If unspecified, the controller watches for machine-api objects across all namespaces.

```

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -4,6 +4,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"time"
 
 	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/cluster-api-provider-gcp/pkg/apis"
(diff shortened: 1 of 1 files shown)


cmd/manager/main.go
```diff
@@ -20,25 +21,56 @@ import (
...
+	watchNamespace := flag.String(
+		"namespace",
+		"",
+		"Namespace that the controller watches to reconcile machine-api objects. If unspecified, the controller watches for machine-api objects across all namespaces.",
+	)
+
+	leaderElectResourceNamespace := flag.String(
+		"leader-elect-resource-namespace",
+		"",
+		"The namespace of resource object that is used for locking during leader election. If unspecified, the controller watches for the namespace currently in-use in the cluster",
```
JoelSpeed, line 39:
```suggestion
		"The namespace of resource object that is used for locking during leader election. If unspecified and running in cluster, defaults to the service account namespace for the controller. Required for leader-election outside of a cluster.",
```

JoelSpeed:
/retitle [OCPCLOUD-492] Implement leader election for gcp provider

PR #86, opened 2020-04-27
Bug 1828288: Deprecated machine.openshift.io/memoryMb annotation in favour of machine.openshift.io/memory

enxebre:
Cluster autoscaler and aws/gcp/azure controllers should drop the "machine.openshift.io/memoryMb" deprecated annotation and use only "machine.openshift.io/memory".

This way controllers are responsible to set the unit format according to https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-memory while the autoscaler should be able to read it appropiately.
This is more robust and practical than the current approach and assumptions.
https://bugzilla.redhat.com/show_bug.cgi?id=1828288

Needs https://github.com/openshift/kubernetes-autoscaler/pull/145
/hold

### DIFF
--- a/pkg/cloud/gcp/actuators/machineset/controller.go
+++ b/pkg/cloud/gcp/actuators/machineset/controller.go
@@ -39,7 +39,7 @@ const (
 	// This is needed by the autoscaler to foresee upcoming capacity when scaling from zero.
 	// https://github.com/openshift/enhancements/pull/186
 	cpuKey    = "machine.openshift.io/vCPU"
-	memoryKey = "machine.openshift.io/memoryMb"
+	memoryKey = "machine.openshift.io/memory"
 )
 
 // Reconciler reconciles machineSets.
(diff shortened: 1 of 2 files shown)


pkg/cloud/gcp/actuators/machineset/controller.go
```diff
@@ -152,7 +152,8 @@ func (r *Reconciler) reconcile(machineSet *machinev1.MachineSet) (ctrl.Result, e
 
 	// TODO: get annotations keys from machine API
 	machineSet.Annotations[cpuKey] = strconv.FormatInt(machineType.GuestCpus, 10)
-	machineSet.Annotations[memoryKey] = strconv.FormatInt(machineType.MemoryMb, 10)
+	// See https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-memory
+	machineSet.Annotations[memoryKey] = fmt.Sprintf("%sM", strconv.FormatInt(machineType.MemoryMb, 10))
```
JoelSpeed, line 156: How do we know this is `M` as opposed to `Mi`? Are there GCP docs that state these values are a particular value? If so we should link those docs as well

elmiko: i was just looking at that as well, according to the link there (https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-memory), the "M" version is just an integer of the bytes whereas the "Mi" version is the power of two unit type. i think in these cases we are better off using "M" since we are consuming the bytes value from the provider. my thought is that we don't want to use the power of two making an assumption that the provider might not have made.

JoelSpeed: From the GCP pricing page https://cloud.google.com/compute/all-pricing

> For Compute Engine, disk size, machine type memory, and network usage are calculated in gigabytes (GB), where 1 GB is 2^30 bytes. This unit of measurement is also known as a gibibyte (GiB).

So they're reported sizes are actually in `GiB`, so I suspect the amount they return in their API is also in binary bytes, therefore `MiB`

enxebre:
https://github.com/openshift/kubernetes-autoscaler/pull/146

PR #87, opened 2020-04-29
apis: support specifying custom hostname for the machines

abhinavdahiya:
GCP allows users to set custom hostname, this esp useful for OS like RHCOS that have limit on the length of the hostname

by default the GCP send DHCP information with the default hostnames as doc in https://cloud.google.com/compute/docs/internal-dns#instance-fully-qualified-domain-names
But these names are quite longer than 64 characters and this cause the RHCOS to fail to add the hostname. Allowing custom hostnames should help setting shorter hostnames.


abhinavdahiya:
/close
```
Error: Error creating instance: googleapi: Error 400: Invalid value for field 'resource.hostname': 'adahiya-2-948lk-b'. Instance hostname is not a valid fully qualified domain name, invalid
```
short hostnames are no possible using custom hostnames :(

PR #88, opened 2020-05-04
Bug 1809345: Add the machine's name as a known NodeInternalDNS

abhinavdahiya:
GCP platform provides search paths to resolve the virtual machine names by default. see https://cloud.google.com/compute/docs/internal-dns#resolv.conf

this also allows nodes in clusters to have node-names matching the virtual machine name instead of the fqdn be because fqdn is too long for RHCOS to accept as hostname
by default

### DIFF
--- a/pkg/cloud/gcp/actuators/machine/reconciler.go
+++ b/pkg/cloud/gcp/actuators/machine/reconciler.go
@@ -197,6 +197,13 @@ func (r *Reconciler) reconcileMachineWithCloudState(failedCondition *v1beta1.GCP
 			Type:    corev1.NodeInternalDNS,
 			Address: fmt.Sprintf("%s.c.%s.internal", r.machine.Name, r.projectID),
 		})
+		// Add the machine's name as a known NodeInternalDNS because GCP platform
+		// provides search paths to resolve those.
+		// https://cloud.google.com/compute/docs/internal-dns#resolv.conf
+		nodeAddresses = append(nodeAddresses, corev1.NodeAddress{
+			Type:    corev1.NodeInternalDNS,
+			Address: r.machine.GetName(),
+		})
 
 		r.machine.Spec.ProviderID = &r.providerID
 		r.machine.Status.Addresses = nodeAddresses


pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -197,6 +197,13 @@ func (r *Reconciler) reconcileMachineWithCloudState(failedCondition *v1beta1.GCP
 			Type:    corev1.NodeInternalDNS,
 			Address: fmt.Sprintf("%s.c.%s.internal", r.machine.Name, r.projectID),
 		})
+		// Add the machine's name as a known NodeInternalDNS because GCP platform
+		// provides search paths to resolve those.
+		// https://cloud.google.com/compute/docs/internal-dns#resolv.conf
+		nodeAddresses = append(nodeAddresses, corev1.NodeAddress{
+			Type:    corev1.NodeInternalDNS,
```
enxebre, line 204: do we need to also include this as `NodeHostName`?

enxebre:
Related PR for ref https://github.com/openshift/machine-config-operator/pull/1711
/approve

michaelgugino:
My only concern here is that the kubelet might be getting node name from the cloud provider (eg, metadata) rather than the output of 'hostname'.  Though, worst case here we have an extra name appended to the machine-object that has no effect.

pkg/cloud/gcp/actuators/machine/reconciler.go
abhinavdahiya: Not really, this VM name is resolvable internal DNS name for the machine on GCP. We can't say what the hostname will be.

abhinavdahiya:
@michaelgugino well what the kubelet can pick a nodename based on various factors, but this change only adds the VM name as one of the internal DNS which is true _orthogonal to nodename, kubelet_ based on https://cloud.google.com/compute/docs/internal-dns#resolv.conf

So on GCP, the kubelet uses the os hostname as node-name. And that can also be change by a flag on kubelet.
But I think for autoapproval we are using the matching or these today (very brittle) due to constraints we have. Could change in the future.

darkmuggle:
I have limited understanding of the dynamics at play. However, in light of openshift/machine-config-operator#1711 this makes sense; 1711 provides for a truncated host name which will be used when the FQDN exceeds 63 characters. 

/lgtm

PR #89, opened 2020-05-13
Bug 1808971: Machine status shows "running" when an instance was terminated

Danil-Grigorev:
Revendor MAO to add unknown state for externally removed machines


JoelSpeed:
@Danil-Grigorev This needs a BZ to unblock it

PR #90, opened 2020-05-18
Bug 1796412: cluster-reader is unable to view machine resources

Danil-Grigorev:
Added cluster-reader rbac for machine CRDs


enxebre:
/hold
thanks @Danil-Grigorev! we don't really need this change.
No manifests are exposed by this repo to payload. Any change must be done into the machine api operator repo

PR #97, opened 2020-06-11
Add support for preemptible instances

alexander-demichev:
Add support for preemptible instances


alexander-demichev:
I haven't tested it yet, in the process of deploying a cluster. I believe this should be all for GCP.

alexander-demichev:
I was able to run a spot instance with these changes

PR #98, opened 2020-06-15
Revendor MAO

alexander-demichev:
Revendor MAO for dropping the requirement to set cluster ID label manually.

openshift/machine-api-operator#608


cmd/manager/main.go
```diff
@@ -64,6 +65,10 @@ func main() {
 		klog.Fatal(err)
 	}
```
enxebre, line 67: can we clarify in the commit or in a code comment why this is needed now?

enxebre: e.g Add configv1 to scheme so the machine controller can fetch the Infrastructure object to set the clusterID

alexander-demichev: done, I updated commit description

enxebre:
thanks Alex!
/approve

go.mod
```diff
@@ -16,8 +16,9 @@ require (
...
 	k8s.io/klog v1.0.0
-	k8s.io/utils v0.0.0-20200327001022-6496210b90e8 // indirect
-	sigs.k8s.io/controller-runtime v0.5.1-0.20200330174416-a11a908d91e0
-	sigs.k8s.io/controller-tools v0.2.9-0.20200331153640-3c5446d407dd
+	sigs.k8s.io/controller-runtime v0.6.0
+	sigs.k8s.io/controller-tools v0.3.0
 	sigs.k8s.io/yaml v1.2.0
 )
+
+replace sigs.k8s.io/cluster-api-provider-aws => github.com/openshift/cluster-api-provider-aws v0.2.1-0.20200612013036-1a97c62077cb
```
enxebre, line 24: why this?

JoelSpeed: I think without this, go.mod goes a bit mad about the dependencies because of what is checked into MAO, it doesn't want to vendor these versions so you have to have an override

https://github.com/openshift/machine-api-operator/blob/ba983a5a8c794d819e3cd7f19afc055a52de7d9e/go.mod#L29-L31

enxebre:
I assume is intentional because the cycle dep that’s coming from the mao API package which brings the provderSpecs now. Also https://github.com/openshift/cluster-api-provider-gcp/pull/98#discussion_r442087817
/hold cancel

enxebre:
https://github.com/openshift/cluster-api-provider-gcp/pull/98#pullrequestreview-433099908 We'll keep introducing changes in a quick cadence during the dev cycle so unless there's anything particular we wanted to pick for this PR I'll get it through as it is.

go.mod
Danil-Grigorev: With latest merges it will require same replaces for azure and gcp, similar to https://github.com/openshift/cluster-api-provider-azure/pull/136/files#diff-37aff102a57d3d7b797f152915a6dc16R30-R34

PR #99, opened 2020-06-22
Add termination handler for spot instances

alexander-demichev:
This PR adds a termination handler for spot instances. It is based on the termination handler that Joel wrote for AWS and has similar logic.

Based on [documentation](https://cloud.google.com/compute/docs/instances/preemptible), spot instances run on GCP for 24 hours, but under some conditions, they might be terminated earlier so polling logic inherited from AWS should be the best option here.

This implementation of the termination handler is simpler than https://github.com/GoogleCloudPlatform/k8s-node-termination-handler, made specifically for machine API, and leverages its ability for draining nodes.

https://cloud.google.com/compute/docs/instances/preemptible

https://github.com/openshift/cluster-api-provider-aws/pull/308

https://cloud.google.com/compute/docs/instances/create-start-preemptible-instance#detecting_if_an_instance_was_preempted

https://github.com/openshift/enhancements/blob/master/enhancements/machine-api/spot-instances.md

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,6 +64,8 @@ sec: # Run security static analysis
 build: ## build binaries
 	$(DOCKER_CMD) go build $(GOGCFLAGS) -o "bin/machine-controller-manager" \
                -ldflags "$(LD_FLAGS)" "$(REPO_PATH)/cmd/manager"
+	$(DOCKER_CMD) go build $(GOGCFLAGS) -o "bin/termination-handler" \
+               -ldflags "$(LD_FLAGS)" "$(REPO_PATH)/cmd/termination-handler"
 
 .PHONY: test-e2e
 test-e2e: ## Run e2e tests
--- a/cmd/termination-handler/main.go
+++ b/cmd/termination-handler/main.go
@@ -1,0 +1,73 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"flag"
+	"fmt"
+	"os"
+	"time"
+
+	"github.com/openshift/cluster-api-provider-gcp/pkg/termination"
+	"github.com/openshift/cluster-api-provider-gcp/pkg/version"
+	"k8s.io/klog"
+	"k8s.io/klog/klogr"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+)
+
+func main() {
+	var printVersion bool
+	flag.BoolVar(&printVersion, "version", false, "print version and exit")
+
+	klog.InitFlags(nil)
+	logger := klogr.New()
+
+	pollIntervalSeconds := flag.Int64("poll-interval-seconds", 5, "interval in seconds at which termination notice endpoint should be checked (Default: 5)")
+	nodeName := flag.String("node-name", "", "name of the node that the termination handler is running on")
+	namespace := flag.String("namespace", "", "namespace that the machine for the node should live in. If unspecified, look for machines across all namespaces.")
+	flag.Set("logtostderr", "true")
+	flag.Parse()
+
+	if printVersion {
+		fmt.Println(version.String)
+		os.Exit(0)
+	}
+
+	// Get a config to talk to the apiserver
+	cfg, err := config.GetConfig()
+	if err != nil {
+		logger.Error(err, "Error getting configuration")
+		return
+	}
+
+	// Get the poll interval as a duration from the `poll-interval-seconds` flag
+	pollInterval := time.Duration(*pollIntervalSeconds) * time.Second
+
+	// Construct a termination handler
+	handler, err := termination.NewHandler(logger, cfg, pollInterval, *namespace, *nodeName)
+	if err != nil {
+		logger.Error(err, "Error constructing termination handler")
+		return
+	}
+
+	// Start the termination handler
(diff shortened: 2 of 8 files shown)


Makefile
```diff
@@ -64,6 +64,8 @@ sec: # Run security static analysis
 build: ## build binaries
 	$(DOCKER_CMD) go build $(GOGCFLAGS) -o "bin/machine-controller-manager" \
                -ldflags "$(LD_FLAGS)" "$(REPO_PATH)/cmd/manager"
+	$(DOCKER_CMD) go build $(GOGCFLAGS) -o "bin/termination-handler" \
+								-ldflags "$(LD_FLAGS)" "$(REPO_PATH)/cmd/termination-handler"
```
JoelSpeed, line 68: Much indent, such space 😜   Possibly we can unindent this a bit
```suggestion
		-ldflags "$(LD_FLAGS)" "$(REPO_PATH)/cmd/termination-handler"
```

cmd/termination-handler/main.go
```diff
@@ -0,0 +1,73 @@
...
+func main() {
+	var printVersion bool
+	flag.BoolVar(&printVersion, "version", false, "print version and exit")
+
+	klog.InitFlags(nil)
+	logger := klogr.New()
+
+	pollIntervalSeconds := flag.Int64("poll-interval-seconds", 5, "interval in seconds at which termination notice endpoint should be checked (Default: 5)")
+	nodeName := flag.String("node-name", "", "name of the node that the termination handler is running on")
+	namespace := flag.String("namespace", "", "namespace that the machine for the node should live in. If unspecified, the look for machines across all namespaces.")
```
JoelSpeed, line 42: Spurious `the` in here that makes this not a valid sentence 😅 
```suggestion
	namespace := flag.String("namespace", "", "namespace that the machine for the node should live in. If unspecified, look for machines across all namespaces.")
```

pkg/termination/termination.go
```diff
@@ -0,0 +1,155 @@
...
+// Run starts the handler and runs the termination logic
+func (h *handler) Run(stop <-chan struct{}) error {
+	ctx, cancel := context.WithCancel(context.Background())
+
+	errs := make(chan error, 1)
+	wg := &sync.WaitGroup{}
+	wg.Add(1)
+	go func() {
+		errs <- h.run(ctx, wg)
+	}()
```
JoelSpeed, line 74: I realise I wrote this for the AWS one, but looking back, I think this would be better to defer the wg.Done() here and not pass it into run, like this:
(When I was reviewing, I was like, "woah, where's the `wg.Done()`")

```suggestion
	go func() {
		defer wg.Done()
		errs <- h.run(ctx)
	}()
```

pkg/termination/termination.go
```diff
@@ -0,0 +1,155 @@
...
+		}
+
+		// Instance not terminated yet
+		logger.V(2).Info("Instance not marked for termination")
+		return false, nil
+	}, ctx.Done()); err != nil {
+		return fmt.Errorf("error polling termination endpoint: %w", err)
+	}
+
+	// Will only get here if the termination endpoint returned 200
```
JoelSpeed, line 131: This comment needs an update please

pkg/termination/termination_test.go
```diff
@@ -0,0 +1,321 @@
...
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/klog/klogr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var notFoundFunc = func(rw http.ResponseWriter, req *http.Request) {
+	rw.Write([]byte("FALSE"))
+}
```
JoelSpeed, line 37: Perhaps we should rename this since it's not a not found anymore, it's a false response, not sure to what though 🤔

Makefile
alexander-demichev: Not my fault, that's VSCode :D

pkg/termination/termination_test.go
alexander-demichev: `notPreemted` ?

Makefile
JoelSpeed: _shakes fist at VSCode_

pkg/termination/termination.go
```diff
@@ -0,0 +1,155 @@
...
+		// Wait for run to stop
+		wg.Wait()
+		return nil
+	case err := <-errs:
+		cancel()
+		return err
+	}
+}
+
+func (h *handler) run(ctx context.Context, wg *sync.WaitGroup) error {
```
JoelSpeed, line 90: The waitgroup here is no longer needed
```suggestion
func (h *handler) run(ctx context.Context) error {
```

pkg/termination/termination_test.go
```diff
@@ -0,0 +1,321 @@
...
+	machinev1 "github.com/openshift/machine-api-operator/pkg/apis/machine/v1beta1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/klog/klogr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var notPreemted = func(rw http.ResponseWriter, req *http.Request) {
```
JoelSpeed, line 35: Missed a `p` in here, name is good otherwise, thanks
```suggestion
var notPreempted = func(rw http.ResponseWriter, req *http.Request) {
```

enxebre:
Can you please include in the description how does this differ and why is it preferred over https://github.com/GoogleCloudPlatform/k8s-node-termination-handler?

I think we should also consider explore a mechanism to proactively decommission machines with similar creation time in a gradual fashion before they expire all at a time. Not a blocker for this PR though.

alexander-demichev:
@enxebre done

config/crds/machine.openshift.io.crd.yaml
```diff
@@ -0,0 +1,120 @@
...
+  name: machines.machine.openshift.io
+spec:
+  additionalPrinterColumns:
+  - JSONPath: .status.providerStatus.instanceId
+    name: Instance
+    description: Instance ID of machine created in AWS
+    type: string
+  - JSONPath: .status.providerStatus.instanceState
+    name: State
+    description: State of the AWS instance
```
enxebre, line 15: aws?

PR #100, opened 2020-07-07
Add spot instance label to machine spec

alexander-demichev:
Add spot instance label to machine spec, we need this label to be passed to the node.

### DIFF
--- a/pkg/cloud/gcp/actuators/machine/reconciler.go
+++ b/pkg/cloud/gcp/actuators/machine/reconciler.go
@@ -248,7 +248,10 @@ func (r *Reconciler) setMachineCloudProviderSpecifics(instance *compute.Instance
 	r.machine.Labels[machinecontroller.MachineAZLabelName] = r.providerSpec.Zone
 
 	if r.providerSpec.Preemptible {
-		r.machine.Labels[machinecontroller.MachineInterruptibleInstanceLabelName] = ""
+		if r.machine.Spec.Labels == nil {
+			r.machine.Spec.Labels = make(map[string]string)
+		}
+		r.machine.Spec.Labels[machinecontroller.MachineInterruptibleInstanceLabelName] = ""
 	}
 }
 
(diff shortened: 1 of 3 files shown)


enxebre:
how could this make it in, didn’t it break its units? https://github.com/openshift/cluster-api-provider-gcp/pull/100/files#diff-6695809a883ef067a0246092fa3d6f22R32

JoelSpeed:
Tests look reasonable, could you possibly copy them over to the other providers too so we have them there as well?

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -248,7 +248,10 @@ func (r *Reconciler) setMachineCloudProviderSpecifics(instance *compute.Instance
 	r.machine.Labels[machinecontroller.MachineAZLabelName] = r.providerSpec.Zone
 
 	if r.providerSpec.Preemptible {
-		r.machine.Labels[machinecontroller.MachineInterruptibleInstanceLabelName] = ""
+		if r.machine.Spec.Labels == nil {
```
enxebre, line 251: tangential to this PR I'd expect this label to be set as `r.machine.Labels` as well so it can be used for filtering. @JoelSpeed thoughts?

pkg/termination/termination.go
```diff
@@ -28,6 +29,7 @@ type Handler interface {
 
 // NewHandler constructs a new Handler
 func NewHandler(logger logr.Logger, cfg *rest.Config, pollInterval time.Duration, namespace, nodeName string) (Handler, error) {
+	machinev1.AddToScheme(scheme.Scheme)
```
enxebre, line 32: can we please create a ticket to add test env coverage for the handler?

pkg/cloud/gcp/actuators/machine/reconciler.go
JoelSpeed: I think it was put here so that it would reflect onto the nodes, we could move it to `r.machine.labels` if those will also get reflected on the node, can't remember quite how that's done, nodelink? (Or we just put it on both here)

alexander-demichev: Yes, node link https://github.com/openshift/machine-api-operator/blob/master/pkg/controller/nodelink/nodelink_controller.go#L239

PR #103, opened 2020-07-15
BUG 1856597: Pass scheme to client creation so that it uses scheme with Machine API

JoelSpeed:
This ensures that we are passing the scheme we are adding the machine api to, into the client. It also fixes the scheme import so that we use the standard client-go scheme instead of the kubectl scheme

### DIFF
--- a/pkg/termination/termination_test.go
+++ b/pkg/termination/termination_test.go
@@ -50,6 +50,8 @@ var _ = Describe("Handler Suite", func() {
 		httpHandler = nil
 		nodeName = "test-node"
 		httpHandler = newMockHTTPHandler(notPreempted)
+		stop = nil
+		errs = nil
 
 		// use NewHandler() instead of manual construction in order to test NewHandler() logic
 		// like checking that machine api is added to scheme
@@ -71,12 +73,10 @@ var _ = Describe("Handler Suite", func() {
 			Expect(err).ToNot(HaveOccurred())
 			h.pollURL = pollURL
 		}
-
-		stop, errs = StartTestHandler(h)
 	})
 
 	AfterEach(func() {
-		if !isClosed(stop) {
+		if stop != nil && !isClosed(stop) {
 			close(stop)
 		}
 		terminationServer.Close()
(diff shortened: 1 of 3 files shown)


enxebre:
Is there a unit test to validate this, i.e the termination handler can access machines?

JoelSpeed:
@enxebre Updated the test suite so that we use a separate scheme for our test code, this should now ensure that the scheme is being registered correctly for the client used by the termination handler

pkg/termination/termination_test.go
```diff
@@ -68,124 +73,128 @@ var _ = Describe("Handler Suite", func() {
...
+				})
 			})
-		})
 
-		Context("and the instance termination notice is not fulfilled", func() {
-			BeforeEach(func() {
-				httpHandler = newMockHTTPHandler(notPreempted)
+			JustBeforeEach(func() {
+				// Ensure the polling logic is excercised in tests
+				for atomic.LoadInt32(&counter) < 4 {
```
elmiko, line 129: i just want to make sure i understand how this counter is working, is this ensuring that the mock termination service is accessed 4 times before the test will continue?

pkg/termination/termination_test.go
```diff
@@ -68,124 +73,128 @@ var _ = Describe("Handler Suite", func() {
...
 
-			It("should not delete the machine", func() {
-				key := client.ObjectKey{Namespace: testMachine.Namespace, Name: testMachine.Name}
-				Consistently(func() error {
-					m := &machinev1.Machine{}
-					return k8sClient.Get(ctx, key, m)
-				}).Should(Succeed())
+			Context("and the poll URL cannot be reached", func() {
+				BeforeEach(func() {
+					h.pollURL = &url.URL{Opaque: "abc#1://localhost"}
```
elmiko, line 184: this is my new favorite url XD

pkg/termination/termination_test.go
JoelSpeed: Yeah that's just so know we've looped a few times before continuing. This change is actually just a whitespace change though, I suggest reviewing this PR with whitespace changes hidden as it's more obvious what I've done here

elmiko: thanks, i'll double check with whitespace

PR #104, opened 2020-07-20
BUG 1858400: [Performance] Lease refresh period for machine-api-controllers is too high, causes heavy writes to etcd at idle

Danil-Grigorev:
The machine-api-controller components are refreshing their lease more
than all other components combined. Bringing this to 90s each, will
decrease etcd writes at idle.

Relevant MAO PR - https://github.com/openshift/machine-api-operator/pull/649


JoelSpeed:
Would it make sense to have a default for this in the MAO repo that can be vendored here? 🤔

PR #105, opened 2020-07-22
Fix minor linting issues

alexander-demichev:
This PR fixes couple minor linting issues

### DIFF
--- a/pkg/cloud/gcp/actuators/util/gcp_credentials.go
+++ b/pkg/cloud/gcp/actuators/util/gcp_credentials.go
@@ -25,7 +25,6 @@ import (
 	"golang.org/x/oauth2/google"
 	apicorev1 "k8s.io/api/core/v1"
 	apimachineryerrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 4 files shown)


pkg/cloud/gcp/actuators/util/gcp_credentials.go
```diff
@@ -25,7 +25,6 @@ import (
 	"golang.org/x/oauth2/google"
 	apicorev1 "k8s.io/api/core/v1"
 	apimachineryerrors "k8s.io/apimachinery/pkg/api/errors"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 	controllerclient "sigs.k8s.io/controller-runtime/pkg/client"
```
JoelSpeed, line 28: Potentially would have kept `client` over `controllerclient`, but I can see that causing clashes, tough decision 🤔

Danil-Grigorev: I would agree that having `client` over `controllerclient` is more preferable, as we primary use the first one anywhere else.

PR #107, opened 2020-07-23
Add tests for actuator events

alexander-demichev:
This PR adds tests for actuator events, but this required couple of other changes.
1. Pass context to machine scope
2. Improve error formatting and make it consistent with other providers
3. Add GCP client builder, similar to what we have in AWS. This allows mocking the client in a cleaner way.
4. Validate machine labels in update()

https://github.com/openshift/machine-api-operator/blob/master/pkg/controller/vsphere/actuator_test.go
https://github.com/openshift/cluster-api-provider-aws/blob/master/pkg/actuators/machine/actuator_test.go

### DIFF
--- a/pkg/cloud/gcp/actuators/machine/actuator.go
+++ b/pkg/cloud/gcp/actuators/machine/actuator.go
@@ -15,7 +15,8 @@ import (
 )
 
 const (
-	scopeFailFmt      = "%s: failed to create scope for machine: %v"
+	scopeFailFmt      = "%s: failed to create scope for machine: %w"
+	reconcilerFailFmt = "%s: reconciler failed to %s machine: %w"
 	createEventAction = "Create"
 	updateEventAction = "Update"
 	deleteEventAction = "Delete"
@@ -45,10 +46,10 @@ func NewActuator(params ActuatorParams) *Actuator {
 // Set corresponding event based on error. It also returns the original error
 // for convenience, so callers can do "return handleMachineError(...)".
 func (a *Actuator) handleMachineError(machine *machinev1.Machine, err error, eventAction string) error {
+	klog.Errorf("%v error: %v", machine.GetName(), err)
 	if eventAction != noEventAction {
 		a.eventRecorder.Eventf(machine, corev1.EventTypeWarning, "Failed"+eventAction, "%v", err)
 	}
-
 	return err
 }
 
@@ -61,12 +62,14 @@ func (a *Actuator) Create(ctx context.Context, machine *machinev1.Machine) error
 		machine:    machine,
 	})
 	if err != nil {
-		return a.handleMachineError(machine, err, createEventAction)
+		fmtErr := fmt.Errorf(scopeFailFmt, machine.GetName(), err)
+		return a.handleMachineError(machine, fmtErr, createEventAction)
 	}
 	if err := newReconciler(scope).create(); err != nil {
 		// Update machine and machine status in case it was modified
 		scope.Close()
-		return a.handleMachineError(machine, err, createEventAction)
+		fmtErr := fmt.Errorf(reconcilerFailFmt, machine.GetName(), createEventAction, err)
+		return a.handleMachineError(machine, fmtErr, createEventAction)
 	}
 	a.eventRecorder.Eventf(machine, corev1.EventTypeNormal, createEventAction, "Created Machine %v", machine.Name)
 	return scope.Close()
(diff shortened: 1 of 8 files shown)


pkg/cloud/gcp/actuators/machine/actuator_test.go
```diff
@@ -1,30 +1,253 @@
...
 func init() {
 	// Add types to scheme
 	machinev1.AddToScheme(scheme.Scheme)
 }
 
-func TestActuatorCreate(t *testing.T) {
-	eventsChannel := make(chan string, 1)
-	recorder := &record.FakeRecorder{
-		Events: eventsChannel,
+func TestMachineEvents(t *testing.T) {
```
Danil-Grigorev, line 30: I think this method should be called `TestActuatorScopeEvents` or just `TestActuatorEvents`

alexander-demichev: 👍

pkg/cloud/gcp/actuators/machine/machine_scope.go
```diff
@@ -163,15 +167,15 @@ func (s *machineScope) PatchMachine() error {
...
-	if err := s.coreClient.Patch(context.Background(), s.machine, s.machineToBePatched); err != nil {
+	if err := s.coreClient.Patch(s.Context, s.machine, s.machineToBePatched); err != nil {
 		klog.Errorf("Failed to patch machine %q: %v", s.machine.GetName(), err)
 		return err
 	}
 
 	s.machine.Status = statusCopy
 
 	// patch status
-	if err := s.coreClient.Status().Patch(context.Background(), s.machine, s.machineToBePatched); err != nil {
```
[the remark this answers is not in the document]
alexander-demichev: I'd leave `s.Context` for better readability.

pkg/cloud/gcp/actuators/machine/actuator.go
```diff
@@ -61,12 +62,14 @@ func (a *Actuator) Create(ctx context.Context, machine *machinev1.Machine) error
 		machine:    machine,
 	})
 	if err != nil {
-		return a.handleMachineError(machine, err, createEventAction)
+		fmtErr := fmt.Errorf(scopeFailFmt, machine.GetName(), err)
```
enxebre, line 70: wouldn't this shadow the error type breaking consumers?

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -168,6 +168,10 @@ func (r *Reconciler) create() error {
 }
 
 func (r *Reconciler) update() error {
+	if err := validateMachine(*r.machine, *r.providerSpec); err != nil {
```
enxebre, line 171: what's the rationale behind introducing this in the PR?

alexander-demichev: We have this validation in other providers, and tests for actuator events cover this case. I'd like to keep these tests consistent among all providers.

pkg/cloud/gcp/actuators/machine/actuator.go
alexander-demichev: In Go 1.13 it shouldn't, https://blog.golang.org/go1.13-errors

pkg/cloud/gcp/actuators/machine/actuator_test.go
```diff
@@ -1,30 +1,253 @@
...
+		MetricsBindAddress: "0",
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	doneMgr := make(chan struct{})
+
+	go func() {
+		g.Expect(mgr.Start(doneMgr)).To(Succeed())
```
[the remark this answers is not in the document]
JoelSpeed: The channel should be closed when the test exits, I don't think there's any issues with the manager not starting right here, apart from the fact we can't detect whether it has started/stopped. If you want to be thorough I would suggest the use of a waitgroup to ensure the goroutine finishes before the tests end

pkg/cloud/gcp/actuators/machine/machine_scope.go
```diff
@@ -18,12 +17,17 @@ import (
 
 // machineScopeParams defines the input parameters used to create a new MachineScope.
 type machineScopeParams struct {
-	coreClient controllerclient.Client
-	machine    *machinev1.Machine
+	context.Context
```
JoelSpeed, line 20: Out of interest, why embed the context rather than have it as an explicit field?

pkg/cloud/gcp/actuators/machine/machine_scope.go
```diff
@@ -18,12 +17,17 @@ import (
...
+
+	coreClient           controllerclient.Client
+	machine              *machinev1.Machine
+	computeClientBuilder computeservice.BuilderFuncType
 }
 
 // machineScope defines a scope defined around a machine and its cluster.
 type machineScope struct {
+	context.Context
```
JoelSpeed, line 29: Same as above, is there a benefit to embedding?

pkg/cloud/gcp/actuators/machine/machine_scope.go
alexander-demichev: I don't think there any benefit in it, but we have the same in vsphere https://github.com/openshift/machine-api-operator/blob/master/pkg/controller/vsphere/machine_scope.go#L33 + I saw a similar thing in upstream https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/master/pkg/context/controller_manager_context.go#L35

PR #108, opened 2020-07-31
Bug 1878170: Add unit tests for machine scope

alexander-demichev:
Add unit tests for machine scope

### DIFF
--- a/pkg/cloud/gcp/actuators/machine/machine_scope_test.go
+++ b/pkg/cloud/gcp/actuators/machine/machine_scope_test.go
@@ -1,0 +1,439 @@
+package machine
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"path/filepath"
+	"testing"
+	"time"
+
+	. "github.com/onsi/gomega"
+	configv1 "github.com/openshift/api/config/v1"
+	gcpv1 "github.com/openshift/cluster-api-provider-gcp/pkg/apis/gcpprovider/v1beta1"
+	computeservice "github.com/openshift/cluster-api-provider-gcp/pkg/cloud/gcp/actuators/services/compute"
+	machinev1 "github.com/openshift/machine-api-operator/pkg/apis/machine/v1beta1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	controllerfake "sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/envtest"
+)
+
+func init() {
+	configv1.AddToScheme(scheme.Scheme)
+	machinev1.AddToScheme(scheme.Scheme)
+}
+
+func TestNewMachineScope(t *testing.T) {
+	g := NewWithT(t)
+
+	userDataSecret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      userDataSecretName,
+			Namespace: defaultNamespaceName,
+		},
+		Data: map[string][]byte{
+			userDataSecretKey: []byte("userDataBlob"),
+		},
+	}
+
+	credentialsSecret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      credentialsSecretName,
+			Namespace: defaultNamespaceName,
+		},
+		Data: map[string][]byte{
+			credentialsSecretKey: []byte("{\"project_id\": \"test\"}"),
+		},
+	}
+
+	invalidCredentialsSecret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      credentialsSecretName,
+			Namespace: defaultNamespaceName,
+		},
+		Data: map[string][]byte{
+			credentialsSecretKey: []byte("1"),
+		},
+	}
+
+	fakeClient := controllerfake.NewFakeClient(userDataSecret, credentialsSecret)
+
+	validProviderSpec, err := gcpv1.RawExtensionFromProviderSpec(&gcpv1.GCPMachineProviderSpec{
+		CredentialsSecret: &corev1.LocalObjectReference{
+			Name: credentialsSecretName,
+		},
+	})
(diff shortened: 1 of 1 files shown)


alexander-demichev:
This PR goes on top of https://github.com/openshift/cluster-api-provider-gcp/pull/107

pkg/cloud/gcp/actuators/machine/machine_scope_test.go
```diff
@@ -0,0 +1,450 @@
...
+				if scope.providerID != expectedProviderID {
+					t.Fatalf("expected to provider id to be %s, got: %s", expectedProviderID, scope.providerID)
+				}
+			}
+		})
+	}
+}
+
+func TestPatchMachine(t *testing.T) {
+	os.Setenv("TEST_ASSET_KUBE_APISERVER", "/tmp/kubebuilder/bin/kube-apiserver")
```
enxebre, line 245: why are hardcoding this here?
If overriding defaults were to be needed this should be handled in the makefile.

pkg/cloud/gcp/actuators/machine/machine_scope_test.go
```diff
@@ -0,0 +1,450 @@
...
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			scope, err := newMachineScope(tc.params)
+
+			if tc.expectedError != nil {
+				if err == nil {
+					t.Fatal("newMachineScope() was expected to return error")
```
enxebre, line 222: Shouldn't this be `Error` rather that `fatal`?

pkg/cloud/gcp/actuators/machine/machine_scope_test.go
```diff
@@ -0,0 +1,450 @@
...
+				if m.Labels["testlabel"] != "test" {
+					return fmt.Errorf("label \"testlabel\" %q not equal expected \"test\"", m.ObjectMeta.Labels["test"])
+				}
+				return nil
+			},
+		},
+		{
+			name: "Test setting phase",
+			mutate: func(m *machinev1.Machine) {
+				m.Status.Phase = &failedPhase
```
enxebre, line 335: fwiw the actuators are not expected to mutate phase

enxebre (changes requested):

enxebre:
thanks! @alexander-demichev this is looking great, I dropped some minor comments

pkg/cloud/gcp/actuators/machine/machine_scope_test.go
alexander-demichev: It's a leftover after I debugged the coded inside IDE

pkg/cloud/gcp/actuators/machine/machine_scope_test.go
alexander-demichev: Yeah, but we are testing `PatchMachine()` behavior here.

pkg/cloud/gcp/actuators/machine/machine_scope_test.go
```diff
@@ -0,0 +1,445 @@
...
+			}
+			gs.Eventually(getMachine, timeout).Should(Succeed())
+
+			machineScope, err := newMachineScope(machineScopeParams{
+				coreClient:           k8sClient,
+				machine:              machine,
+				Context:              ctx,
+				computeClientBuilder: computeservice.MockBuilderFuncType,
+			})
+
```
enxebre, line 419: nit: remove extra line

enxebre:
/approve
@JoelSpeed @Danil-Grigorev PTAL

pkg/cloud/gcp/actuators/machine/machine_scope_test.go
```diff
@@ -0,0 +1,445 @@
...
+	"context"
+	"errors"
+	"fmt"
+	"path/filepath"
+	"testing"
+	"time"
+
+	. "github.com/onsi/gomega"
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/cluster-api-provider-gcp/pkg/apis/gcpprovider/v1beta1"
```
JoelSpeed, line 13: Nit: Would be better to alias this when we import so that we don't have a random `v1beta1` amongst all the other APIs, `gcpv1` might be a better alias. We've also imported it elsewhere as `gcp`

pkg/cloud/gcp/actuators/machine/machine_scope_test.go
```diff
@@ -0,0 +1,445 @@
...
+				return fmt.Errorf("phase is nil or not equal expected \"Failed\"")
+			},
+		},
+		{
+			name: "Test setting provider status",
+			mutate: func(m *machinev1.Machine) {
+				instanceID := "123"
+				instanceState := "running"
+				providerStatus.InstanceID = &instanceID
+				providerStatus.InstanceState = &instanceState
```
JoelSpeed, line 345: This is using a "global", shouldn't this be setting the providerStatus on the particular machine itself? Otherwise it could interfere with other tests?

pkg/cloud/gcp/actuators/machine/machine_scope_test.go
```diff
@@ -0,0 +1,445 @@
...
+				machine:              machine,
+				Context:              ctx,
+				computeClientBuilder: computeservice.MockBuilderFuncType,
+			})
+
+			gs.Expect(err).ToNot(HaveOccurred())
+
+			tc.mutate(machineScope.machine)
+
+			machineScope.providerStatus = providerStatus
```
JoelSpeed, line 424: I would expect this to be done in mutate if it's being done at all

pkg/cloud/gcp/actuators/machine/machine_scope_test.go
```diff
@@ -0,0 +1,445 @@
...
+				return tc.expect(machine)
+			}
+			gs.Eventually(checkExpectation, timeout).Should(Succeed())
+
+			// Check that resource version doesn't change if we call patchMachine() again
+			machineResourceVersion := machine.ResourceVersion
+
+			gs.Expect(machineScope.PatchMachine()).To(Succeed())
+			gs.Eventually(getMachine, timeout).Should(Succeed())
+			gs.Expect(machine.ResourceVersion).To(Equal(machineResourceVersion))
```
JoelSpeed, line 442: You should replace this with a consistently, we need to make sure that it doesn't change for a period, not that it hasn't changed the first time we call get which is what's currently happening here I think

JoelSpeed (commented):
Looks good, left a few suggestions.

Might be nice if the whole suite used gomega rather than half of it, WDYT?

JoelSpeed (approved):
/lgtm

PR #110, opened 2020-08-03
Disable keep alives on termination handler connection

JoelSpeed:
The GCP metadata service keeps the connnection alive, this means that if we try to intercept calls to the metadata service after the connection is established, we have to restart the process.

We use this interception technique during the E2E suite to verify that the termination handler will respond approriately to a termination event if observed.

### DIFF
--- a/pkg/termination/termination.go
+++ b/pkg/termination/termination.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"io/ioutil"
+	"net"
 	"net/http"
 	"net/url"
 	"sync"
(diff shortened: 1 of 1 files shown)


pkg/termination/termination.go
```diff
@@ -90,6 +91,28 @@ func (h *handler) Run(stop <-chan struct{}) error {
...
+			}).DialContext,
+			ForceAttemptHTTP2:     true,
+			MaxIdleConns:          100,
+			IdleConnTimeout:       90 * time.Second,
+			TLSHandshakeTimeout:   10 * time.Second,
+			ExpectContinueTimeout: 1 * time.Second,
+
+			// By disabling keep alives, this allows us to intercept metadata traffic during E2E testing
+			// and allows us to simulate a termination event.
+			DisableKeepAlives: true,
```
michaelgugino, line 112: This seems like a large compromise just to make tests work.

Can you link to what we're doing in the e2e tests for reference?

JoelSpeed: > This seems like a large compromise just to make tests work.

Agreed, though I can't really think of a better way to allow this test to work. I've tried killing the connection using `ss` from the privileged container produced by the test, but to no avail. We could disable the test/delete it, but that's also not great.

For reference it is the test that was introduced in: https://github.com/openshift/cluster-api-actuator-pkg/pull/171, do you think this should be linked in the code somewhere?

JoelSpeed:
/hold

This may not have resolved the issue, need to do some manual verification to work out what's going on

JoelSpeed:
Closing in favour of https://github.com/openshift/cluster-api-actuator-pkg/pull/185

PR #111, opened 2020-08-10
 Make termination handler subscription based

Danil-Grigorev:
Use google compute metadata service to watch termination event and utilize the callback method, which would trigger machine deletion.

This PR brings no significant changes to existing polling algorithm, delegates this task to the server. The overall polling URL does not change, but the response time to the metadata changes is quicker.

This is important, as this termination handler is our only guarantee, that the node won't freeze with all scheduled pods, once the spot instance is terminated. As the termination handler thanks to `DaemonSet` is running on the same instance, which should be watched for preemption, it's sudden stop will be devastating to the overall cluster health.

This PR is following recommended practices mentioned in https://github.com/openshift/cluster-api-provider-gcp/issues/112

### DIFF
--- a/cmd/termination-handler/main.go
+++ b/cmd/termination-handler/main.go
@@ -20,7 +20,6 @@ import (
 	"flag"
 	"fmt"
 	"os"
-	"time"
 
 	"github.com/openshift/cluster-api-provider-gcp/pkg/termination"
 	"github.com/openshift/cluster-api-provider-gcp/pkg/version"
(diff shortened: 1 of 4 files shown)


cmd/termination-handler/main.go
```diff
@@ -37,7 +36,7 @@ func main() {
 	klog.InitFlags(nil)
 	logger := klogr.New()
 
-	pollIntervalSeconds := flag.Int64("poll-interval-seconds", 5, "interval in seconds at which termination notice endpoint should be checked (Default: 5)")
+	_ = flag.Int64("poll-interval-seconds", 5, "interval in seconds at which termination notice endpoint should be checked (Default: 5)")
```
elmiko, line 39: does this flag get registered as a side effect of calling `flag.Int64` ?

i'm just curious if we don't need to save it, can we just drop it?

JoelSpeed:
/hold

I would prefer we don't do this for a couple of reasons:
1) This means we lose any logging that is currently existing, as a user it is useful to have a regular ping or something to know that the handler is actually doing something and not just hanging (which this currently sort of appears to do)
2) It complicates the logic and reduces the similarity between this and the other two handlers. One of the benefits of the termination handlers at present is that all three are nearly identical, this means that they are easy to maintain and understand. This PR adds complexity which makes it harder to understand and I don't really see that it adds all that much value.

I'm willing to have my mind changed but I think we would need a very good reason to justify this change, why is this better than the current approach?

michaelgugino:
I agree with the hold, this isn't targeting a bug, so not a priority.

cmd/termination-handler/main.go
Danil-Grigorev: We could, but this is the part which is manually set in `sync.go` by the MAO as one of the mandatory arguments. That will require a patch on its own to rely on default values. It would be preferable to find alternatives to that PR in other providers.

PR #113, opened 2020-08-11
BUG 1867262: Support networks shared from a different project

JoelSpeed:
This allows users to specify an alternative project for the network that the VM is created within. This allows for shared VPCs.

I have yet to test this in a real cluster, I want to verify this works before allowing it to merge, hence holding. This shouldn't block the review process though.

/hold


JoelSpeed:
Update, I have tested this in an openshift 4.6 cluster built via UPI into a shared VPC. I had to also add the `projectID` to the subnetwork address to get this to work, but I can confirm it does now work and I managed to successfully create a machine using the Machine API.

The only addition required for this was for me to manually allow the Machine API google service account created for the cluster to have the "Compute Network User" role assigned to it in the Host (network host) project. The specific required permission was `compute.subnetworks.use`.

PR #115, opened 2020-08-12
BUG 1858400: [Performance] Lease refresh period for machine-api-controllers is too high, causes heavy writes to etcd at idle

Danil-Grigorev:
Reduce default lease retry rate on `30s` which will prevent from heavy writes into etcd at idle, and constrain renew deadline on `90s`.

Inspired by https://github.com/openshift/cloud-credential-operator/pull/231
Depends on https://github.com/openshift/machine-api-operator/pull/675

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -24,6 +24,13 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 )
 
+// The default durations for the leader electrion operations.
+var (
+	leaseDuration = 120 * time.Second
+	renewDealine  = 110 * time.Second
+	retryPeriod   = 20 * time.Second
+)
+
 func main() {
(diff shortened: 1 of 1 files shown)


JoelSpeed:
/lgtm

There was some outstanding conversation on the BZ, we can follow up with a second PR if needed

elmiko:
> There was some outstanding conversation on the BZ, we can follow up with a second PR if needed

i thought we had agreed to proceed with these patches and then add the releaseOnCancel stuff in a later patch?

iiuc, these PRs that extend the time at least help the situation.

JoelSpeed:
> i thought we had agreed to proceed with these patches and then add the releaseOnCancel stuff in a later patch?

Yeah sorry, it was based on our conversation that I gave the LGTM, we can follow up with releaseOnCancel later I believe

PR #122, opened 2020-09-28
BUG 1875598: Ensure the Virtual Machine provider state is set to Unknown when Failed

JoelSpeed:
Revendor to include the bug fix from https://github.com/openshift/machine-api-operator/pull/696


michaelgugino (approved):
/lgtm

PR #132, opened 2020-10-20
Enable Machine API to support Customer Managed Keys

JoelSpeed:
This enables users to specify the required information for creating VMs using KMS keys from GCP. Primarily this will allow use of Customer Managed Encryption Keys.

This is as of yet untested, I just wanted to throw together a PR to get a very bare implementation to review.


alexander-demichev (approved):
/approve

pkg/apis/gcpprovider/v1beta1/gcpmachineproviderconfig_types.go
```diff
@@ -75,3 +76,24 @@ type GCPServiceAccount struct {
 	Email  string   `json:"email"`
 	Scopes []string `json:"scopes"`
 }
+
+// GCPEncryptionKeyReference describes the encryptionKey to use for a disk's encryption.
+type GCPEncryptionKeyReference struct {
```
michaelgugino, line 81: Do we need this extra wrapper struct?  It might be useful if there are other types of keys we might need to support in the future, but I don't know if those exist?

JoelSpeed: You've hit the nail on the head 😅  You can also specify other options normally (such as providing a raw key) but I didn't want to implement those in a first pass. To make sure we can support those in a future I added the wrapper struct

For reference [this](https://godoc.org/google.golang.org/api/compute/v1#CustomerEncryptionKey) is the struct on the compute module, extras there are ServiceAccount and RawKey. I'm trying to test this now and we might actually need the ServiceAccount anyway 🤔

JoelSpeed:
I've now tested this and successfully created a VM using a CMEK. I've added the service account option as well as I believe the customer will want to use this.

I discovered during testing that the default compute engine service account is used for the decryption which is not what I had expected

elmiko (commented):
this generally makes sense to me

Danil-Grigorev (approved):
/lgtm

staebler (commented):
These additions look good to me.

JoelSpeed:
/hold cancel

Lets get this merged an unblock the installer implementation

PR #134, opened 2020-11-11
[OCPCLOUD-1044] Bug 1902157: Update termination handler to use node conditions

alexander-demichev:
Switch from deleting the Machine to marking the Node with a Terminating condition.
This will integrate with an MHC to ensure that the Machine is deleted.
This should be possible using the Node's credentials and as such, the termination handler will not need its own credentials and extra permissions.

Requires openshift/machine-api-operator#627 before it will work

https://github.com/openshift/cluster-api-provider-aws/pull/332

### DIFF
--- a/pkg/termination/termination.go
+++ b/pkg/termination/termination.go
@@ -10,7 +10,8 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
-	machinev1 "github.com/openshift/machine-api-operator/pkg/apis/machine/v1beta1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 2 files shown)


pkg/termination/termination.go
```diff
@@ -131,27 +128,76 @@ func (h *handler) run(ctx context.Context) error {
 	}
 
 	// Will only get here if the termination endpoint returned FALSE
-	logger.V(1).Info("Instance marked for termination, deleting Machine")
-	if err := h.client.Delete(ctx, machine); err != nil {
-		return fmt.Errorf("error deleting machine: %w", err)
+	logger.V(1).Info("Instance marked for termination, marking Machine for deletion")
+	if err := h.markNodeForDeletion(ctx); err != nil {
+		return fmt.Errorf("error marking machine: %v", err)
```
JoelSpeed, line 133: This is saying to mark the machine, needs to say to mark the node

elmiko (commented):
looks mostly good to me, just need to fix the logs that Joel mentioned

JoelSpeed:
/retitle [OCPCLOUD-1044] Bug 1902157: Update termination handler to use node conditions

Danil-Grigorev (approved):
/lgtm

PR #138, opened 2020-12-01
Vendor: downgrade gcp sdk to v0.57.0

michaelgugino:
This commit ensures we match the installer's version of
cloud.google.com/go to enable compability with terraform.


JoelSpeed:
I've verified that this change doesn't break anything. The `google.golang.org/api` package was updated to include a new field from version `v0.15.0` to `v0.33.0`. This PR reverts this package to `v0.29.0` but that still contains the new field that was required.

/approve

elmiko (approved):
/lgtm

PR #139, opened 2020-12-03
Bug 1903424: Ensure MAO dependency is up to date to include newest metrics

JoelSpeed:
This updates MAO to the latest version. This should be done periodically to include new updates to the Machine controller code. In particular, this brings in some new metrics that should allow us to debug slow booting machines in the future.


Danil-Grigorev:
/approve

Makes sense to me, needs a revendor though.

michaelgugino (approved):
/lgtm

PR #140, opened 2020-12-15
Bug 1907286: Ensure Machine is marked interruptible as well as Node

JoelSpeed:
To enable the termination handler MHC to work, we need to label the Machine as well as the Node as interruptible.


elmiko (approved):
/lgtm

PR #145, opened 2021-01-25
Try to mark node for up to 30 seconds when termination notice observed

JoelSpeed:
Currently, if any error is raised during the process of marking the node, the container exits. This causes the pod to go into a crash loop backoff and therefore, we pretty rapidly exceed the 30 second timeout period we have to start draining a node. This is an attempt to minimise this risk by trying every second for up to 30 seconds to mark the node. If that fails, then we timeout and give up.

This will hopefully reduce failures such as those [caused by DNS](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-api-actuator-pkg/205/pull-ci-openshift-cluster-api-actuator-pkg-master-e2e-gcp-operator/1353694838301659136/artifacts/e2e-gcp-operator/test/termination-handler.log) in test runs.

### DIFF
--- a/pkg/termination/termination.go
+++ b/pkg/termination/termination.go
@@ -129,7 +129,18 @@ func (h *handler) run(ctx context.Context) error {
 
 	// Will only get here if the termination endpoint returned FALSE
 	logger.V(1).Info("Instance marked for termination, marking Node for deletion")
-	if err := h.markNodeForDeletion(ctx); err != nil {
+
+	// Try every second to mark the node for termination up to a 30 second timeout.
+	// This should help to prevent intermittent errors and ensure we don't end up in crash loop backoff.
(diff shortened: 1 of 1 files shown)


alexander-demichev (approved):
/approve

michaelgugino:
/lgtm

Do we have any indication why DNS is sometimes failing?  I would think that should be quite stable in a CI cluster.

JoelSpeed:
Realised I was breaking out of the loop by returning the error, so have updated to log the error and not return.

> Do we have any indication why DNS is sometimes failing? I would think that should be quite stable in a CI cluster.

No, I'm going to raise it to the test platform folks today

elmiko (approved):
/lgtm

PR #150, opened 2021-03-04
Bug 1934021: Ensure response body is closed when we are finished with the request

JoelSpeed:
When creating an http request in go, this sets two goroutines in the background running. If the body is not closed properly, these goroutines are leaked. This leads to a gradual increase in memory used by the application over time. We must ensure we always close response bodies

### DIFF
--- a/pkg/termination/termination.go
+++ b/pkg/termination/termination.go
@@ -104,6 +104,9 @@ func (h *handler) run(ctx context.Context) error {
 		req.Header.Add("Metadata-Flavor", "Google")
 
 		resp, err := http.DefaultClient.Do(req)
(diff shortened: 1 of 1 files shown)


elmiko (approved):
/approve

alexander-demichev (approved):
/lgtm

JoelSpeed:
/override ci/prow/e2e-gcp-operator

The test failures here are completely unrelated to this code, I'm confident that this is working as expected and that it is safe to override

JoelSpeed:
/override ci/prow/e2e-gcp-operator

The flake is unrelated to this PR

PR #153, opened 2021-03-19
add in-container vendor

SamuelStuchly:
Add option to run make vendor in container like other targets, to avoid issues with go module caching depending on what you already have on your machine.

### DIFF
--- a/hack/go-mod.sh
+++ b/hack/go-mod.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+go mod tidy
+go mod vendor
+go mod verify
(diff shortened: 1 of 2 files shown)


elmiko (approved):
/approve

hack/go-mod.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+go mod tidy
+go mod vendor
+go mod verify
```
JoelSpeed, line 4: Nit: Should have a new line at the end of this file ideally

PR #158, opened 2021-04-07
Update GCP default disk image to include ignition v3

JoelSpeed:
The webhook test started failing consistently for GCP because the default image that was being used does not support ignition v3. Update to a newer image (as used in the installer) so that the tests pass again


elmiko (approved):
/approve
/lgtm

PR #159, opened 2021-04-16
Bug 1948719: update k8s dependencies to 1.21

elmiko:
This change updates the go.mod and related vendor files to the
kubernetes 1.21 baseline. This was achieved by updating some of the
entries in go.mod and then running make vendor.

The updates to go.mod were as follows:
```
github.com/openshift/machine-api-operator master
```

Then ran `make vendor` to produce output.


elmiko:
this commit is needed by https://github.com/openshift/cluster-api-actuator-pkg/pull/209

PR #168, opened 2021-08-05
Bug 1990447: Create event only if the machine was modified

Fedosin:
Now gcp actuator always creates an event after Update, even though the change might not have happened.

To prevent creating fake events we compare machine resource version before and after update, and create an event only if they are different.

### DIFF
--- a/pkg/cloud/gcp/actuators/machine/actuator.go
+++ b/pkg/cloud/gcp/actuators/machine/actuator.go
@@ -117,8 +117,21 @@ func (a *Actuator) Update(ctx context.Context, machine *machinev1.Machine) error
 		fmtErr := fmt.Errorf(reconcilerFailFmt, machine.GetName(), updateEventAction, err)
 		return a.handleMachineError(machine, fmtErr, updateEventAction)
 	}
-	a.eventRecorder.Eventf(machine, corev1.EventTypeNormal, updateEventAction, "Updated Machine %v", machine.Name)
-	return scope.Close()
+
+	previousResourceVersion := scope.machine.ResourceVersion
+
+	if err := scope.Close(); err != nil {
+		return err
+	}
+
+	currentResourceVersion := scope.machine.ResourceVersion
+
+	// Create event only if machine object was modified
+	if previousResourceVersion != currentResourceVersion {
+		a.eventRecorder.Eventf(machine, corev1.EventTypeNormal, updateEventAction, "Updated Machine %v", machine.Name)
+	}
+
+	return nil
 }
 
 func (a *Actuator) Delete(ctx context.Context, machine *machinev1.Machine) error {
--- a/pkg/cloud/gcp/actuators/machine/actuator_test.go
+++ b/pkg/cloud/gcp/actuators/machine/actuator_test.go
@@ -233,8 +233,16 @@ func TestActuatorEvents(t *testing.T) {
 
 			// Create the machine
 			gs.Expect(k8sClient.Create(context.Background(), machine)).To(Succeed())
+
+			// Make sure the machine and its event are deleted when the test ends
 			defer func() {
 				gs.Expect(k8sClient.Delete(context.Background(), machine)).To(Succeed())
+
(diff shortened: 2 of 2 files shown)


pkg/cloud/gcp/actuators/machine/actuator.go
```diff
@@ -72,12 +72,19 @@ func (a *Actuator) Create(ctx context.Context, machine *machinev1.Machine) error
...
+			return err
+		}
 		fmtErr := fmt.Errorf(reconcilerFailFmt, machine.GetName(), createEventAction, err)
 		return a.handleMachineError(machine, fmtErr, createEventAction)
 	}
 	a.eventRecorder.Eventf(machine, corev1.EventTypeNormal, createEventAction, "Created Machine %v", machine.Name)
-	return scope.Close()
+
+	if err := scope.Close(); err != nil {
+		return err
```
JoelSpeed, line 84: Might be worth adding some extra context to this error with a `fmt.Errorf`

pkg/cloud/gcp/actuators/machine/actuator_test.go
```diff
@@ -215,8 +215,6 @@ func TestActuatorEvents(t *testing.T) {
 
 	for _, tc := range cases {
 		t.Run(tc.name, func(t *testing.T) {
-			gs := NewWithT(t)
```
JoelSpeed, line 218: We use a separate `gomega` so that when the test fails, it's only the subtest, not the whole test. Please revert removing this and the renaming of the previous `g` declaration as well

pkg/cloud/gcp/actuators/machine/actuator_test.go
```diff
@@ -230,13 +230,18 @@ func TestActuatorEvents(t *testing.T) {
...
-
-			// Create the machine
-			gs.Expect(k8sClient.Create(context.Background(), machine)).To(Succeed())
+			eventList := &v1.EventList{}
+			// Make sure the machine and its event are deleted when the test ends
 			defer func() {
 				gs.Expect(k8sClient.Delete(context.Background(), machine)).To(Succeed())
+				for i := range eventList.Items {
+					gs.Expect(k8sClient.Delete(context.Background(), &eventList.Items[i])).To(Succeed())
+				}
```
JoelSpeed, line 239: To avoid any sharing of the eventList, and to make sure we delete events even when there's an earlier failure, why not do a list in here to make sure we have an up to date list
```suggestion
				
				eventList := &v1.EventList{}
				gs.Expect(k8sClient.List(context.Background(), eventList).To(Succeed())
				for i := range eventList.Items {
					gs.Expect(k8sClient.Delete(context.Background(), &eventList.Items[i])).To(Succeed())
				}
```

JoelSpeed (commented):
To avoid potential for leaking events, when cleaning up, we should make sure to do a fresh list, please include my suggestion (and you can move the other eventList declaration back to where it was) and I think we are good to go

pkg/cloud/gcp/actuators/machine/actuator.go
```diff
@@ -72,12 +72,19 @@ func (a *Actuator) Create(ctx context.Context, machine *machinev1.Machine) error
 	}
 	if err := newReconciler(scope).create(); err != nil {
 		// Update machine and machine status in case it was modified
-		scope.Close()
+		if err := scope.Close(); err != nil {
+			return err
```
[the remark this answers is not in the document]
Fedosin: I think this change is not really related to the bugfix, so I removed it from here. we can implement it in a separate pr.

pkg/cloud/gcp/actuators/machine/actuator_test.go
Fedosin: Yes, makes sense to me! thanks for the suggestion.

JoelSpeed:
/approve

Thanks Mike

alexander-demichev (approved):
/lgtm

PR #172, opened 2021-08-31
[OCPCLOUD-812] Enable support for instances with GPUs on GCP

SamuelStuchly:
Support instances of "A2" family as well as all five GPUs available with "N1" instances. Regular and preeimptible instances.

Does not support GPUs for graphics workloads.


pkg/apis/gcpprovider/v1beta1/gcpmachineproviderconfig_types.go
```diff
@@ -104,3 +108,9 @@ type GCPKMSKeyReference struct {
 	// Location is the GCP location in which the Key Ring exists.
 	Location string `json:"location"`
 }
+
+// GCPAcceleratorConfig describes type and count of accelerator cards attached to the instance on GCP.
+type GCPAcceleratorConfig struct {
+	AcceleratorCount int64  `json:"acceleratorCount,omitempty"`
+	AcceleratorType  string `json:"acceleratorType,omitempty"`
```
JoelSpeed, line 115: Please add descriptions to these fields as these will end up in `oc describe`

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -39,6 +38,91 @@ func newReconciler(scope *machineScope) *Reconciler {
...
+func containsString(sli []string, str string) bool {
+	for _, elem := range sli {
+		if elem == str {
+			return true
+		}
+	}
+	return false
+}
+
+func (r *Reconciler) checkQuota(machineFamily int64) error {
```
JoelSpeed, line 61: Maybe this would be a more obvious as a name for this variable as to what this number really is?
```suggestion
func (r *Reconciler) checkQuota(machineTypeAcceleratorCount int64) error {
```

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -39,6 +38,91 @@ func newReconciler(scope *machineScope) *Reconciler {
...
+	region, err := r.computeService.RegionGet(r.projectID, r.providerSpec.Region)
+	if err != nil {
+		return fmt.Errorf("failed to get region via compute service: %v", err)
+	}
+	quotas := region.Quotas
+	var guestAccelerators = []*v1beta1.GCPAcceleratorConfig{}
+	if machineFamily != 0 {
+		guestAccelerators = append(guestAccelerators, &v1beta1.GCPAcceleratorConfig{AcceleratorType: "nvidia-tesla-a100", AcceleratorCount: machineFamily})
+	} else {
```
JoelSpeed, line 70: This could use a comment, something along the lines of `When the machine type has associated accelerator instances, these will be A100s. Additional guest accelerators are not allowed so ignore the providerSpec GuestAccelerators.`

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -39,6 +38,91 @@ func newReconciler(scope *machineScope) *Reconciler {
...
+		}
+	} else if containsString(n1MachineFamily, machineType) {
+		// n1 family machine
+		if err := r.checkQuota(0); err != nil {
+			return err
+		} else {
+			return nil
+		}
+	} else {
+		return nil
```
JoelSpeed, line 122: What happens if a user specifies a guest accelerator and they aren't using an A2 or an N1 instance type?

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -39,6 +38,91 @@ func newReconciler(scope *machineScope) *Reconciler {
...
+				return machinecontroller.InvalidMachineConfiguration(fmt.Sprintf("No quota found. Metric: %s.", metric))
+			}
+		}
+	}
+	return nil
+}
+
+func (r *Reconciler) validateGuestAccelerators() error {
+
+	a2MachineFamily, n1MachineFamily := r.computeService.MachineTypesList(r.providerSpec.ProjectID, r.providerSpec.Zone, r.Context)
```
JoelSpeed, line 105: Should we make sure we only make this API call when it's definitely required, ie guest accelerators are listed or it's an A2 type?

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -56,10 +140,25 @@ func (r *Reconciler) create() error {
...
+			Preemptible:       r.providerSpec.Preemptible,
+			AutomaticRestart:  r.providerSpec.AutomaticRestart,
+			OnHostMaintenance: r.providerSpec.OnHostMaintenance,
 		},
 	}
 
+	var guestAccelerators = []*compute.AcceleratorConfig{}
+	for index, ga := range r.providerSpec.GuestAccelerators {
+		guestAccelerators = append(guestAccelerators, &compute.AcceleratorConfig{
+			AcceleratorType:  fmt.Sprintf("zones/%s/acceleratorTypes/%s", zone, r.providerSpec.GuestAccelerators[index].AcceleratorType),
```
JoelSpeed, line 152: Please move the format here to a const at the top of the file

pkg/cloud/gcp/actuators/services/compute/computeservice.go
```diff
@@ -101,3 +108,32 @@ func (c *computeService) TargetPoolsRemoveInstance(project string, region string
 func (c *computeService) MachineTypesGet(project string, zone string, machineType string) (*compute.MachineType, error) {
 	return c.service.MachineTypes.Get(project, zone, machineType).Do()
 }
+
+func (c *computeService) MachineTypesList(project string, zone string, ctx context.Context) (map[string]int64, []string) {
```
JoelSpeed, line 112: Perhaps we should name this to identify that it is for MachineTypes that support guest accelerators?

JoelSpeed (commented):
This looks really good, I've left a few comments inline but for the most part they are stylistic or naming nits, good work!

pkg/cloud/gcp/actuators/machine/reconciler.go
SamuelStuchly: googleapi Error gets is returned later when trying [`InstancesInsert()`](https://github.com/openshift/cluster-api-provider-gcp/blob/master/pkg/cloud/gcp/actuators/machine/reconciler.go#L147) looking like this:
`googleapi: Error 400: [<machineType>, <AcceleratorType>] features are not compatible for creating instance., badRequest`

JoelSpeed: I think we can probably return an error here then instead, as we know that this isn't supported. We can write a more informative error for the user I think

SamuelStuchly: Something like this ? `InvalidMachineConfiguration(fmt.Sprintf("MachineType %s does not support accelerators.",machinteType))`
Or do we want to be more informative and display accelerator as googleapi error does and maybe specify that only A1 and N1 support accelerators ?

JoelSpeed: I think that's good, I'd add a note on the end as well saying `Only A1 and N1 machine type families support guest acceleartors`

pkg/apis/gcpprovider/v1beta1/gcpmachineproviderconfig_types.go
```diff
@@ -104,3 +108,11 @@ type GCPKMSKeyReference struct {
 	// Location is the GCP location in which the Key Ring exists.
 	Location string `json:"location"`
 }
+
+// GCPAcceleratorConfig describes type and count of accelerator cards attached to the instance on GCP.
+type GCPAcceleratorConfig struct {
+	// AcceleratorCount is number of AcceleratorType accelerators (GPUs) to be attached to an instance
+	AcceleratorCount int64 `json:"acceleratorCount,omitempty"`
+	// AcceleratorType is the type of accelerator (GPU) to be attached to an instance
```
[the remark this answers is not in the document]
SamuelStuchly: Makes sense since it is for the `oc describe`

Btw supported accelerators are listed in a structure https://github.com/openshift/cluster-api-provider-gcp/pull/172/files#diff-359c93fbc08f569c3abf79881f0b426e681c46a5e76765d05e921f16fdb3bf99R44

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -50,16 +144,31 @@ func (r *Reconciler) create() error {
...
+			Preemptible:       r.providerSpec.Preemptible,
+			AutomaticRestart:  r.providerSpec.AutomaticRestart,
+			OnHostMaintenance: r.providerSpec.OnHostMaintenance,
 		},
 	}
 
+	var guestAccelerators = []*compute.AcceleratorConfig{}
+	for index, ga := range r.providerSpec.GuestAccelerators {
+		guestAccelerators = append(guestAccelerators, &compute.AcceleratorConfig{
+			AcceleratorType:  fmt.Sprintf(acceleratorTypeFmt, zone, r.providerSpec.GuestAccelerators[index].AcceleratorType),
```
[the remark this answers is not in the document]
SamuelStuchly: My mistake. I assumed since `r.providerSpec.GuestAccelerators` is slice, it might have more then one `AcceleratorConfig`s in it. However it cannot have more than one. 

Same logic applied [here](https://github.com/openshift/cluster-api-provider-gcp/pull/172/files#diff-359c93fbc08f569c3abf79881f0b426e681c46a5e76765d05e921f16fdb3bf99R79).

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -39,6 +40,91 @@ func newReconciler(scope *machineScope) *Reconciler {
...
+	var guestAccelerators = []*v1beta1.GCPAcceleratorConfig{}
+	// When the machine type has associated accelerator instances (A2 machine family), accelerators will be nvidia-tesla-A100s.
+	// Additional guest accelerators are not allowed so ignore the providerSpec GuestAccelerators.
+	if machineTypeAcceleratorCount != 0 {
+		guestAccelerators = append(guestAccelerators, &v1beta1.GCPAcceleratorConfig{AcceleratorType: "nvidia-tesla-a100", AcceleratorCount: machineTypeAcceleratorCount})
+	} else {
+		guestAccelerators = r.providerSpec.GuestAccelerators
+	}
+	// validate zone and then quota
+	accelerator := guestAccelerators[0] // guestAccelerators slice cannot be longer than 1
```
JoelSpeed, line 79: Why is this limited to 1? Is this a GCP limitation, if so, where is this written in the docs?

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -39,6 +40,99 @@ func newReconciler(scope *machineScope) *Reconciler {
...
+			if i == len(quotas)-1 {
+				return machinecontroller.InvalidMachineConfiguration(fmt.Sprintf("No quota found. Metric: %s.", metric))
+			}
+		}
+	}
+	return nil
+}
+
+func (r *Reconciler) validateGuestAccelerators() error {
+	if len(r.providerSpec.GuestAccelerators) != 0 || strings.HasPrefix(r.providerSpec.MachineType, "a2-") {
```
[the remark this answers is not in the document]
JoelSpeed: Logically that means that with the current version, a regular machineType with 0 accelerators is allowed to go through and make the API call, I think you should double check that as it is today, and update the if statement here, perhaps it should be...
```suggestion
	if (len(r.providerSpec.GuestAccelerators) != 0 && strings.HasPrefix(r.providerSpec.MachineType, "n1-") || strings.HasPrefix(r.providerSpec.MachineType, "a2-") {
```

pkg/cloud/gcp/actuators/machine/reconciler.go
SamuelStuchly: I did not find it the docs, however gcp web console does not even have an option for attaching more then 1 type of GPU. And I tried to create an instance with 2 different guest accelerators through the api, and got an googleapi error saying  guestaccelerators is of size 2 which is too long.

SamuelStuchly:
\hold

JoelSpeed (commented):
/approve

Thanks Sam!

elmiko (commented):
nice work Sam
/lgtm

PR #173, opened 2021-09-10
Updat golang image to version 1.16 inside Makefile

SamuelStuchly:
Current master branch is no longer able to build with golang 1.15, therefore it would be nice to update Makefile to use 1.16 image.


elmiko (commented):
/lgtm

thanks Sam!

PR #180, opened 2021-11-12
 Update golang to version 1.17

elmiko:
This change updates the go.mod, makefile, dockerfiles, and the vendor directory to use golang 1.17 for building. it also adds a minor convenience option to the makefile that allows a user to override the build image. It also adds a fix to the Makefile for building with containers.


elmiko:
update
* ran `make fmt` to update build tags

alexander-demichev (approved):
/lgtm

PR #191, opened 2023-03-06
OCPBUGS-2135: Add cloud provider label to instances

RadekManak:
Adds "kubernetes-io-cluster-\<clusterName\>: owned" label to created instances to enable openshift (un)installer to find and delete instances created by capg. 

This label (with some differences due to character limitations) is added by other providers, but is commented out on GCP for some reason.

NameKubernetesAWSCloudProviderPrefix = "kubernetes.io/cluster/" on AWS
NameKubernetesAzureCloudProviderPrefix = "kubernetes.io_cluster_" on Azure

Alternatively, we could update the installer to also look for "capg-cluster-\<clusterName\>" label that is already being added.

### DIFF
--- a/api/v1beta1/labels.go
+++ b/api/v1beta1/labels.go
@@ -37,11 +37,11 @@ func (in Labels) HasOwned(cluster string) bool {
 	return ok && ResourceLifecycle(value) == ResourceLifecycleOwned
 }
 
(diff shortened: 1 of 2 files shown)


JoelSpeed:
Have you looked into why its currently commented out? I'd like to understand that before we proceed

RadekManak:
It was commented out since the first commit that added it.

PR #220, opened 2024-02-15
OCPBUGS-28331: add test directory to snyk excludes

elmiko:
we do not package or release anything from the test directory and it is safe to exclude from our snyk scans.

### DIFF
--- a/.snyk
+++ b/.snyk
@@ -5,3 +5,4 @@ exclude:
   global:
     - "vendor/**"
     - "**/vendor/**"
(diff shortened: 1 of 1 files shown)


damdo:
@elmiko this will need the `UPSTREAM: <carry>` notation

elmiko:
updated commit message

JoelSpeed:
/override ci/prow/e2e-gcp

This PR doesn't affect anything that's in the payload

PR #222, opened 2024-03-20
OCPBUGS-30586: fix e2e tests on release branches

damdo:
Fix e2e tests on release branches by running cluster-capi-operator tests off of the matching release branch.


damdo:
This change doesn't involve payload code, only tooling. So the failures are not related. Hence overriding

/override ci/prow/e2e-gcp-serial ci/prow/e2e-gcp

PR #224, opened 2024-05-03
OCPBUGS-33170: Container fallback to logs on error

damdo:
In 4.17, we are enforcing this value for all OpenShift containers via tests, link in bug comment. This helps with debugging to ensure we have logs available.

Ref: https://github.com/openshift/cluster-capi-operator/pull/172


JoelSpeed:
/override ci/prow/e2e-gcp-serial

The job is running the default featureset, which does not yet include anything from this repository. Looking at the failure, it appears to be a networking (OVN) related issue and nothing to do with this change

PR #238, opened 2025-02-24
NO-JIRA: Update manifests generator tooling

RadekManak:
Adds make target that updates the manifest generation tool to latest version. This make target will be run by lifecycle hook to ensure the tool is always up-to-date.

It removes the downstream vendored version of kustomize. This was version was made redundant when we started running manifests-gen that uses its kustomize as a library. Keeping it around was only breaking things when features are deprecated.

Always running the latest version of manifest-gen will allow us to be ahead of the deprecations made to kustomize. When one repository uses kustomize feature that manifest-gen does not have yet. We only need to update manifest-gen in cluster-capi-operator and the rebases will pick it up.


damdo:
Unrelated failure

/override ci/prow/e2e-gcp-serial

PR #243, opened 2025-06-02
OCPBUGS-56962: [release-4.19] Update manifests to remove dependency on admissionregistration v1beta1

theobarberbany:
This PR updates manifests gen and then regenerates the manifests to pick up switch from v1beta1 to v1 for the admissionregistration group on the ValidatingAdmissionPolicy{Binding}s we use


JoelSpeed:
/override ci/prow/security

Unrelated, won't fix issues

PR #270, opened 2026-02-24
OCPBUGS-78283: [release-4.13] hermetic migration 4.13

fgallott:
this is needed for the hermetic migration in konflux, to fix the following error:

```
Error: PackageRejected: The content of the vendor directory is not consistent with go.mod. Please check the logs for more details.
  Please try running `go mod vendor` and committing the changes.
  Note that you may need to `git add --force` ignored files in the vendor/ dir.
```

same as https://github.com/openshift/cluster-api-provider-gcp/pull/269

PR #271, opened 2026-02-24
OCPBUGS-78284: [release-4.12] hermetic migration 4.12

fgallott:
same as https://github.com/openshift/cluster-api-provider-gcp/pull/270


nrb:
Will become valid after #270 merges

PR #292, opened 2026-05-06
NO-JIRA: manifests-gen: scope provider webhooks to capi namespace

damdo:
Bumps `manifests-gen` to include webhook namespace scoping, which adds a `namespaceSelector` to all provider webhooks restricting them to the `openshift-cluster-api` namespace.

This prevents webhooks from intercepting requests in other namespaces, which would fail on unsupported platforms where the webhook server is not running.

## Summary by CodeRabbit

* **Chores**
  * Updated multiple webhook configurations to restrict admission control operations exclusively to the openshift-cluster-api namespace. This ensures all mutation and validation webhooks for GCP-related cluster resources, machine templates, managed clusters, and managed machine pools operate only within the designated cluster API namespace.
  * Updated manifests-gen dependency to a newer version for improved compatibility.


damdo (commented):
/approve

